PR #19, opened 2019-09-05
Add basic demos of launching TWAs.

PEConn:
Launching a TWA without using LauncherActivity is really ergonomic, except for the case of using a custom referrer.

I wanted to get these demos up so we had a vague use case for before dealing with https://github.com/GoogleChrome/android-browser-helper/issues/13 .


PEConn:
Andre, Pavel - looking at the code it looks like the appropriate place for this to do is TrustedWebActivityIntentBuilder alongside setAdditionalTrustedOrigins, would you agree?

We could add it in TwaLauncher instead as a stop gap for a quicker turn around.

Or Andre suggested we could have some sort of "customize Intent callback" that TwaLauncher takes and then executes just before launching the Intent. That would be really powerful (in both the good and bad way).

pshmakov:
"customize Intent callback" sounds good. Alternatively, a less powerful option may work - receive a Bundle of additional extras to put into Intent.

demo/build.gradle
```diff
@@ -33,11 +33,19 @@ android {
...
+    }
 }
 
 dependencies {
+    api 'androidx.browser:browser:1.2.0-alpha07'
+    implementation project(path: ':androidbrowserhelper')
+
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation 'androidx.appcompat:appcompat:1.0.2'
+    implementation 'androidx.constraintlayout:constraintlayout:1.1.3'
```
pshmakov: Nit: unused.

andreban:
> Or Andre suggested we could have some sort of "customize Intent callback" that TwaLauncher takes and then executes just before launching the Intent. That would be really powerful (in both the good and bad way).

The already has this power (eg: writing their own `TwaLauncher`) or just launching their own Intent. It's a matter of how easy we are making it for them to change the Intent.

PEConn:
> The already has this power (eg: writing their own TwaLauncher) or just launching their own Intent. It's a matter of how easy we are making it for them to change the Intent.

It's also a matter of the public API. What they can do today is messy, but that's fine because it uses the building blocks we provide. If we add a "customize intent callback" that then becomes part of TwaLauncher's API and people will start relying on it as well - especially if we move the Referrer stuff into AndroidX, then we'll have this method people will have been using be redundant.

The other aspect is letting developers shoot themselves in the foot - the EXTRA_REFERRER has a fairly specific format. If we give them raw access to it, they may get it wrong. I think we should provide an interface like Bundle and then construct it ourselves (as Pavel suggested).

However these do seem like orthogonal points - maybe we should create a createReferrerFromKeyValuePairs method and let them call it in the customizeIntentCallback method.

andreban:
I agree that the higher level API for setting the referral, etc is the best approach. It's more obvious to developers and less likely for them to shoot themselves on the foot.

Regarding the intent callback, it provides a good path for developers who have use-cases we don't know about yet and would need access to the Intent.

demo/src/main/java/com/google/androidbrowserhelper/launchtwa/LaunchTwaActivity.java
```diff
@@ -0,0 +1,95 @@
...
+
+    public void launchWithCustomColors(View view) {
+        TrustedWebActivityIntentBuilder builder = new TrustedWebActivityIntentBuilder(LAUNCH_URI)
+                        .setNavigationBarColor(Color.RED)
+                        .setToolbarColor(Color.BLUE);
+
+        new TwaLauncher(this).launch(builder, null, null);
+    }
+
+    public void launcherWithMultipleOrigins(View view) {
```
willyau: this is great!

PEConn:
Hey Andre, added the licenses.

PR #21, opened 2019-09-06
Remove buildToolsVersion from library build.gradle

### LINKED ISSUE #20: Android Browser Helper should not declare a buildToolsVersion
Declaring a buildToolsVersion in the library's Gradle file leads to a warning in Android Studio, since this is no longer recommended. The specific buildToolsVersion should be removed from the Gradle file.

andreban:
- When using Gradle 3.0 or above, it defaults to use a buildTool
  version specified by the Gradle plugin. buildToolsVersion should
  only be used in order to override the default version
- Reference: https://developer.android.com/studio/releases/build-tools
- Fixes #20

### DIFF
--- a/androidbrowserhelper/build.gradle
+++ b/androidbrowserhelper/build.gradle
@@ -18,8 +18,6 @@ apply plugin: 'com.android.library'
 
 android {
     compileSdkVersion 28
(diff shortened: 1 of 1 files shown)


PEConn:
Thanks for taking care of this Andre.

PR #23, opened 2019-09-09
[DO NOT MERGE] Test Bintray Configuration

andreban:
- Uploads binary, source code and javadocs
- Based on docs here: https://github.com/bintray/bintray-examples/tree/master/gradle-bintray-plugin-examples

@PEConn did some investigation on setting-up Bintray. This is not final, but got it somewhat working and uploading the library.

### DIFF
--- a/.idea/gradle.xml
+++ b/.idea/gradle.xml
@@ -8,7 +8,15 @@
         </compositeConfiguration>
         <option name="distributionType" value="DEFAULT_WRAPPED" />
         <option name="externalProjectPath" value="$PROJECT_DIR$" />
+        <option name="modules">
+          <set>
+            <option value="$PROJECT_DIR$" />
+            <option value="$PROJECT_DIR$/androidbrowserhelper" />
+            <option value="$PROJECT_DIR$/demo" />
+          </set>
+        </option>
         <option name="resolveModulePerSourceSet" value="false" />
+        <option name="testRunner" value="PLATFORM" />
       </GradleProjectSettings>
     </option>
   </component>
--- a/androidbrowserhelper/publish.gradle
+++ b/androidbrowserhelper/publish.gradle
@@ -1,0 +1,90 @@
+apply plugin: 'maven-publish'
+apply plugin: 'com.jfrog.bintray'
+
+version '0.1.0-alpha4'
+group 'me.bandarra'
+
+
+Properties properties = new Properties()
+properties.load(project.rootProject.file('local.properties')
+        .newDataInputStream())
+
+// build.gradle
+task sourcesJar(type: Jar) {
+    from android.sourceSets.main.java.source
+    classifier = 'sources'
+}
+task javadoc(type: Javadoc) {
+    source = android.sourceSets.main.java.source
+    ext.androidJar = "${android.sdkDirectory}/platforms/${android.compileSdkVersion}/android.jar"
+    classpath += (project.files(android.getBootClasspath().join(File.pathSeparator))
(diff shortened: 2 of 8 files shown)


.idea/gradle.xml
```diff
@@ -8,7 +8,15 @@
         </compositeConfiguration>
         <option name="distributionType" value="DEFAULT_WRAPPED" />
         <option name="externalProjectPath" value="$PROJECT_DIR$" />
+        <option name="modules">
```
PEConn: I'm assuming this was an accident - why is this file not caught by the .gitignore?

demo/build.gradle
```diff
@@ -1,44 +1,89 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
  */
 
 apply plugin: 'com.android.application'
 
+def twaManifest = [
```
PEConn: I think you suggested it on my pull request about adding demos - we should probably have multiple smaller demos so that the manifest is not everything pushed together.

We can't have multiple android apps in a single module can we? I guess we'll just have to have multiple demo modules.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed Aug 07 09:18:46 BST 2019
+#Mon Sep 09 13:15:19 BST 2019
```
PEConn: Again, assuming accidental add.

demo/build.gradle
```diff
@@ -1,44 +1,89 @@
...
+        applicationId: 'org.chromium.twa.svgomg',
```
PEConn: So will this eventually replace svgomg-twa?

androidbrowserhelper/publish.gradle
```diff
@@ -0,0 +1,90 @@
...
+    pkg {
+        repo = 'android-browser-helper'
+        name = 'androidbrowserhelper'
+        userOrg = 'andreban'
+        desc = 'A Companion Library for Web On Android'
+        licenses = ['Apache-2.0']
+        vcsUrl = 'https://github.com/andreban/android-browser-helper'
+        githubRepo = 'andreban/android-browser-helper'
+        githubReleaseNotesFile = 'README.md'
+//        dryRun = true
```
PEConn: Courageous :-P

androidbrowserhelper/publish.gradle
```diff
@@ -0,0 +1,90 @@
...
+    key = properties.getProperty("bintray.apikey")
+    pkg {
+        repo = 'android-browser-helper'
+        name = 'androidbrowserhelper'
+        userOrg = 'andreban'
+        desc = 'A Companion Library for Web On Android'
+        licenses = ['Apache-2.0']
+        vcsUrl = 'https://github.com/andreban/android-browser-helper'
+        githubRepo = 'andreban/android-browser-helper'
+        githubReleaseNotesFile = 'README.md'
```
PEConn: So the release notes for all releases will be checked into this repo somewhere? (That seems like a good thing, just wondering).

andreban:
Hey @PEConn, thanks for taking a look. I don't think we should use this PR as the final state for the bintray config. I just didn't want to lose this until we have a proper account. Once we have, we can take the pieces that matter here into another proper PR

Issue #8, opened 2019-09-04
Add demos


PEConn:
Add some examples to the demos module. Potentially copying from [custom-tabs-client](https://github.com/GoogleChrome/custom-tabs-client).

PEConn:
And add docs to build them.

andreban:
It seems it's worth renaming the demos folder to `demo` and having multiple modules inside it (eg: CCT, TWAs, etc). WDYT?

andreban:
We have a `demos` folder with a couple of projects. We can create issues for more demos as needed. Closing this in favor of tracking specific demos

PR #29, opened 2019-10-01
Renames `demo` to `demos`

andreban:
- Adds a `demos` directoru
- Moves code inside the `demo` project to `demos/custom-launch-activity`
- Updates gradle files to point to the new folder structure.

Improves on #8


PEConn:
I think the name custom-launch-activity is a bit weird (originally I thought it was to do with Custom Tabs) - maybe something like 'launching-twas' would be better?

Also, we could probably get rid of DemoListActivity and make the demo just start off with LaunchTwaActivity.

However, I'm not too invested in either idea, so LGTM!

andreban:
Maybe `custom-twa-launcher`? `launching-twas` has the issue that other demos will also launch TWAs (eg: an SVGOMG-like demo).

Regarding refactoring the demo, I want to do that in a different PR, as moving code around around combined with actual code changes make it annoying to review.

PR #31, opened 2019-10-02
Multi domain manifest

andreban:


demos/twa-multi-domain/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,54 @@
...
+        <meta-data
+            android:name="asset_statements"
+            android:resource="@string/asset_statements" />
+
+        <activity android:name="com.google.androidbrowserhelper.trusted.LauncherActivity"
+            android:label="@string/app_name">
+
+            <!-- Additional origins to be validated with Digital AssetLinks are listed under
+                this meta-tag, which points to a string-array in strings.xml -->
+            <meta-data android:name="android.support.customtabs.trusted.DOMAINS_TO_VALIDATE"
```
PEConn: This needs to be updated right (the DOMAINS_TO_VALIDATE)?

demos/twa-multi-domain/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,54 @@
...
+                android:value="@array/additional_trusted_origins" />
+
+            <meta-data android:name="android.support.customtabs.trusted.ADDITIONAL_TRUSTED_ORIGINS"
```
PEConn: Now I'm confused with what you're doing... Aren't you meant to have ADDITIONAL_TRUSTED_ORIGINS set to @array/additional_trusted_origins ?

demos/twa-multi-domain/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+
+dependencies {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+    api 'androidx.browser:browser:1.2.0-alpha07'
+    implementation project(path: ':androidbrowserhelper')
+    implementation 'androidx.appcompat:appcompat:1.0.2'
+
+    testImplementation 'junit:junit:4.12'
+    androidTestImplementation 'androidx.test:runner:1.2.0'
+    androidTestImplementation 'androidx.test.espresso:espresso-core:3.2.0'
```
PEConn: You don't need the lines about testing dependencies.

demos/twa-multi-domain/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
+}
+
+dependencies {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+    api 'androidx.browser:browser:1.2.0-alpha07'
+    implementation project(path: ':androidbrowserhelper')
+    implementation 'androidx.appcompat:appcompat:1.0.2'
```
PEConn: I don't think you need this dependency.

demos/twa-multi-domain/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
+}
+
+dependencies {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+    api 'androidx.browser:browser:1.2.0-alpha07'
```
PEConn: I don't think you need this dependency either.

demos/twa-multi-domain/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+        versionCode 1
+        versionName "1.0"
+
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+    }
+
+    buildTypes {
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
```
PEConn: You could probably get rid of these lines and proguard-rules.pro if we want to keep things simple.

demos/twa-multi-domain/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+
+
+    defaultConfig {
+        applicationId "com.google.browser.examples.twa_multi_domain"
+        minSdkVersion 26
+        targetSdkVersion 29
+        versionCode 1
+        versionName "1.0"
+
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
```
PEConn: We have no tests! Can get rid of this.

demos/twa-multi-domain/src/main/AndroidManifest.xml
andreban: Yep - I changed the wrong key when renaming. This one was supposed to be LAUNCH_URL

demos/twa-multi-domain/src/main/AndroidManifest.xml
andreban: Yes.

demos/twa-multi-domain/build.gradle
andreban: done

demos/twa-multi-domain/build.gradle
andreban: done.

demos/twa-multi-domain/build.gradle
andreban: done.

demos/twa-multi-domain/build.gradle
andreban: Kind of - It does build and run, but Android Studio marks some things as invalid in the UI. I'm, removing, as I think this is more like an Android Studio issue and shouldn't be here anyway.

demos/twa-multi-domain/build.gradle
andreban: done

PR #34, opened 2019-10-16
Generate Maven Artifacts

andreban:
- Modified `androidbrowserhelper/build.gradle` to generate the
   Maven artifacts required for publishing the library.

### DIFF
--- a/androidbrowserhelper/build.gradle
+++ b/androidbrowserhelper/build.gradle
@@ -15,6 +15,7 @@
  */
 
 apply plugin: 'com.android.library'
+apply plugin: 'maven-publish'
 
 android {
     compileSdkVersion 28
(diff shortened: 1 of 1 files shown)


androidbrowserhelper/build.gradle
```diff
@@ -61,3 +62,41 @@ dependencies {
...
+                }
+            }
+
+            artifact("$buildDir/outputs/aar/androidbrowserhelper-release.aar")
+
+            groupId 'com.google.androidbrowserhelper'
+            artifactId 'androidbrowserhelper'
+            version '0.1.0-alpha1'
+
+            pom.withXml {
```
PEConn: Could you add a comment over why this is necessary (eg link to the Stack Overflow)?

androidbrowserhelper/build.gradle
```diff
@@ -61,3 +62,41 @@ dependencies {
...
+                def dependenciesNode = asNode().appendNode('dependencies')
+
+                // Iterate over the implementation dependencies (we don't want the test ones), adding a <dependency> node for each
```
PEConn: nit: line wrapping.

andreban:
@PEConn PTAL

androidbrowserhelper/build.gradle
andreban: Wrapped the line, but added a link that is large :(

Issue #36, opened 2019-11-07
Library crashes when used with androidx.browser latest version


homer-jay:
This library crashes when used with [androidx.browser](https://developer.android.com/jetpack/androidx/releases/browser) latest version (as of today, v1.2.0-alpha09).

### Steps to reproduce
1.  Configure the `app/build.gradle` with the latest versions of both androidx.browser and android-browser-helper:
```
api 'androidx.browser:browser:1.2.0-alpha09'
implementation 'com.google.androidbrowserhelper:androidbrowserhelper:0.1.0-alpha1'
```

2. Launch a Chrome Custom Tab like in the demo example:
```
TrustedWebActivityIntentBuilder builder = new TrustedWebActivityIntentBuilder(Uri.parse("https://www.google.com"))
	.setNavigationBarColor(Color.RED)
	.setToolbarColor(Color.RED);

new TwaLauncher(this).launch(builder, null, null);
```

### Expected result
The tab should open without any error. This is the case if we use androidx.browser version 1.2.0-alpha08.

### Actual result
An exception is thrown. Stack trace:
```
java.lang.NoSuchMethodError: No virtual method getUrl()Landroid/net/Uri; in class Landroidx/browser/trusted/TrustedWebActivityIntentBuilder; or its super classes (declaration of 'androidx.browser.trusted.TrustedWebActivityIntentBuilder' appears in /data/app/org.example.customtabsdemo-1/base.apk)
        at com.google.androidbrowserhelper.trusted.TwaLauncher.launchCct(TwaLauncher.java:139)
        at com.google.androidbrowserhelper.trusted.TwaLauncher.launch(TwaLauncher.java:127)
        at org.example.customtabsdemo.android.ui.AppListActivity.showWebBrowser(AppListActivity.java:193)
        at org.example.customtabsdemo.android.ui.AppListActivity.openApplication(AppListActivity.java:171)
        at org.example.customtabsdemo.android.ui.AppListActivity.access$600(AppListActivity.java:40)
        at org.example.customtabsdemo.android.ui.AppListActivity$1.onRowClicked(AppListActivity.java:116)
        at org.example.customtabsdemo.android.ui.adapters.AppDescriptorAdapter.lambda$onBindViewHolder$1$AppDescriptorAdapter(AppDescriptorAdapter.java:92)
[...]

PR #37, opened 2019-11-07
Update README with how to use the library

andreban:
- Even though the library is not stable yet, it already has a
  version available on the Google Maven. Since Android Browser
  Helper is the only way to use TWAs with AndroidX libraries,
  adding instructions for developers who want to do so, along
  with a warning on why it isn't stable yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,23 @@ developer will find themselves requiring, for example:
 * Code for choosing an appropriate Custom Tabs provider.
 * Creating an Activity to launch the browser's site settings for a TWA.
 
+## Adding Android Browser Helper to an Android project
+
+Android Browser helper is available on the Google Maven. To use it, modify your application's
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,6 +11,23 @@ developer will find themselves requiring, for example:
...
+    //...
+    implementation 'com.google.androidbrowserhelper:androidbrowserhelper:0.1.0-alpha1'
+}
+
+``` 
+
+**Important:** Since Android Browser Helper depends on a version of `androidx.browser` that is not
+yet stable, the library is still marked as alpha. This will be updated once all the libraries
+it depends on are stable.
+
```
PEConn: Could you add a comment warning about https://github.com/GoogleChrome/android-browser-helper/issues/36 ?

PEConn: Thanks!

PR #39, opened 2019-11-25
Create Basic TWA demo.

PEConn:
Hey Andre, take a look!


demos/twa-basic/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+        applicationId "com.google.browser.examples.twa_basic"
+        minSdkVersion 26
+        targetSdkVersion 29
+        versionCode 1
+        versionName "1.0"
+    }
+
+    buildTypes {
+        release {
+            minifyEnabled false
```
andreban: We may want to default `minifyEnabled` to `true`. Also, do we plan on adding rules to proguard? Otherwise, we may want to delete it.

PEConn: Sure, enabled minify and got rid of the proguard rules file.

PR #41, opened 2019-11-27
Provide an Intent to the browser to focus the TWA.

PEConn:
Hey,

This is the long awaited counterpart to https://chromium-review.googlesource.com/c/chromium/src/+/1849682


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/FocusActivity.java
```diff
@@ -0,0 +1,45 @@
...
+public class FocusActivity extends AppCompatActivity {
+    // This value should be moved into androidx.browser.
+    private static final String EXTRA_FOCUS_INTENT =
+            "androidx.browser.customtabs.extra.FOCUS_INTENT";
+
+    public static void addToIntent(Intent containerIntent, Context context) {
+        Intent focusIntent = new Intent(context, FocusActivity.class);
+
+        // This class may not be included in app's manifest, don't add it in that case.
+        if (focusIntent.resolveActivityInfo(context.getPackageManager(), 0) == null) return;
```
pshmakov: This is on the startup path, and calling package manager has some cost...
I think the best we can do is to cache the result in a static field.

PEConn:
Closed in favour of https://github.com/GoogleChrome/android-browser-helper/pull/53

PR #45, opened 2019-12-10
Adds README.md for twa-multi-domain

andreban:
- Documents  how to add more origins when validating a TWA.


andreban:
It won't. I wonder if this is actually desirable in all cases.

PR #46, opened 2019-12-12
Creates a Demo for a WebView Fallback

andreban:
- Modifies TwaLauncher and LauncherActivity to allow customising the fallback strategy
- Adds an Activity with a naive WebView implementation.
- Further PRs should extend the functionality of the WebView and harden the implementation.


demos/twa-webview-fallback/src/main/res/xml/filepaths.xml
```diff
@@ -0,0 +1,16 @@
...
+     Unless required by applicable law or agreed to in writing, software
+     distributed under the License is distributed on an "AS IS" BASIS,
+     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+     See the License for the specific language governing permissions and
+     limitations under the License.
+-->
+
+<paths>
+    <files-path path="twa_splash/" name="twa_splash" />
+</paths>
```
PEConn: nit: lack of newline

demos/twa-webview-fallback/src/main/res/layout/activity_web_view_fallback.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context=".WebViewFallbackActivity">
+
+</androidx.constraintlayout.widget.ConstraintLayout>
```
PEConn: nit: lack of newline.

PEConn: Also if this class is to remain empty, you can just use `/>` on the end of line 7 and get rid of line 9.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -261,4 +280,11 @@ public void onServiceDisconnected(ComponentName componentName) {
...
         }
     }
+
+    public interface FallbackStrategy {
+        void launch(Context context,
+                    TrustedWebActivityIntentBuilder twaBuilder,
+                    @Nullable String providerPackage,
+                    @Nullable Runnable completionCallback);
+    }
 }
```
PEConn: nit: add newline

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -261,4 +280,11 @@ public void onServiceDisconnected(ComponentName componentName) {
             mSession = null;
         }
     }
+
+    public interface FallbackStrategy {
```
PEConn: nit: could you move the interface to the top (after private variables).

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -219,6 +224,20 @@ public void destroy() {
         mDestroyed = true;
     }
 
+    public static final FallbackStrategy cctFalbackStrategy =
```
PEConn: Could you move this to the top (with the definition of Fallback Strategy)?

Also, static final variables should be BLOCK_CAPS.

demos/twa-webview-fallback/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.google.browser.examples.twawebviewfallback">
+
+    <uses-permission android:name="android.permission.INTERNET"/>
```
PEConn: Do we need this line?

demos/twa-webview-fallback/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="utf-8"?>
```
PEConn: You need licenses and such here (and probably other files too).

I would be nice if we could highlight (maybe just in a comment at the top) what's special about this AndroidManifest (that it's the standard one plus WebViewFallbackActivity).

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -219,6 +224,20 @@ public void destroy() {
...
+        // CustomTabsIntent will fall back to launching the Browser if there are no Custom Tabs
+        // providers installed.
+        CustomTabsIntent intent = twaBuilder.buildCustomTabsIntent();
+        if (providerPackage != null) {
+            intent.intent.setPackage(providerPackage);
+        }
+        intent.launchUrl(context, twaBuilder.getUrl());
+        if (completionCallback != null) {
+            completionCallback.run();
+        }
```
PEConn: I think we should move this part out of the fallbackStrategy - we don't want users to forget about calling it.

So take a FallbackStrategy (which is now only a `(context, twaBuilder, providerPackage) -> ...`) and then maybe within launch we wrap it in another callback that calls it and then completionCallback.

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
```diff
@@ -0,0 +1,106 @@
...
+import android.webkit.WebView;
+import android.webkit.WebViewClient;
+import android.widget.FrameLayout;
+import android.widget.Toast;
+
+public class WebViewFallbackActivity extends AppCompatActivity {
+    public static final String KEY_LAUNCH_URI =
+            "com.google.browser.examples.twawebviewfallback.WebViewFallbackActivity.KEY_LAUNCH_URL";
+
+    private Uri launchUrl;
```
PEConn: nit: variable should be called mLaunchUrl;

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        this.launchUrl = this.getIntent().getParcelableExtra(KEY_LAUNCH_URI);
+
+        FrameLayout webViewContainer = new FrameLayout(this);
```
PEConn: Do we need the FrameLayout? Could we just have a WebView and nothing else?

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
```diff
@@ -0,0 +1,106 @@
...
+                // URL, in this example.
+                Toast.makeText(view.getContext(), "Recovering from crash",
+                        Toast.LENGTH_LONG).show();
+                webView.loadUrl(launchUrl.toString());
+                return true;
+            }
+
+            @Override
+            public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request) {
+                Uri launchUrl = WebViewFallbackActivity.this.launchUrl;
```
PEConn: This will break in the event that someone creates a TWA that launches to a site off the verified origin. I don't know why they'd do that, so maybe we could just say they're holding it wrong at that point.

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
```diff
@@ -0,0 +1,106 @@
...
+                Uri navigationUrl = request.getUrl();
+
+                // If the user is navigation to a different origin, use CCT to handle the navigation
+                if (!launchUrl.getScheme().equals(navigationUrl.getScheme())
```
PEConn: Probably outside the scope of this initial CL, but I'd be very interested to see if the OAuth dance would work. WebView Fallback -> CCT -> Custom Scheme -> Launch TWA -> WebView Fallback.

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
```diff
@@ -0,0 +1,106 @@
...
+        webSettings.setJavaScriptEnabled(true);
+
+        FrameLayout.LayoutParams layoutParams = new FrameLayout.LayoutParams(
+                FrameLayout.LayoutParams.MATCH_PARENT, FrameLayout.LayoutParams.MATCH_PARENT);
+
+        setContentView(webViewContainer,layoutParams);
+        webView.loadUrl(launchUrl.toString());
+    }
+
+    private WebViewClient createWebViewClient() {
```
PEConn: Are we happy with the WebView client having references to the Activity? We probably are, but just want to make sure there's not a memory leak or something.

demos/twa-webview-fallback/src/main/res/xml/filepaths.xml
andreban: done

demos/twa-webview-fallback/src/main/res/layout/activity_web_view_fallback.xml
andreban: Removed the entire layout, as I'm creating the WebView programatically.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
andreban: done

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
andreban: done

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
andreban: done

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
andreban: done

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
andreban: You are right -  I though we needed some kind of ViewGroup at the root, but it seems the root itself is a ViewGroup.

andreban: done

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
andreban: Actually, we are not validating origins at all in the fallback. It seems worth doing it in another PR.

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
andreban: +1 - I don't think it's going to work right now. My current goal is to add a naive implementation that won't handle more complex dances (and then iterate)

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
andreban: The WebViewClient is attached to the WebView, which is attached to the Activity. When the Activity is GC'd, those will be GC'd with it. Having said that, we can refactor this and pass the launchUrl as a parameter when constructing the WebViewClient.

PEConn:
Apart from the `<uses-permission android:name="android.permission.INTERNET"/>` in the manifest, this all LGTM.

andreban:
Without the INTERNET permission, the WebView will not be able to load content :(

andreban:
I stand corrected - I had an issue previously and thought it was fixed by adding the permission. It does work without it.

PR #48, opened 2020-01-02
Makes android-browser-helper compatible with androidx.browser-1.2.0

andreban:
- Updates Key names to versions compatible with `androidx.browser`
- Implements a `TokenStore` with `SharedPreferencesTokenStore`
- Handles the deprecation of `PackageInfo.versionCode`
- Renames `DelegationService` to `TrustedWebActivityService` and
  implements `getTokenStore`
- Updates `twa-basic` and `twa-webview-fallback` to use
  `com.google.androidbrowserhelper.helper.TrustedWebActivityService`
  instead of `androidx.browser.truster.TrustedWebActivityService`

**WARNING**: This is a breaking change, due to developers having to
replace the androidx.browser TrustedWebActivityService with the
android-browser-helper TrustedWebActivityService.

Closes #48

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ChromeLegacyUtils.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ChromeLegacyUtils.java
@@ -148,6 +148,8 @@ public class ChromeLegacyUtils {
     static int getVersionCode(PackageManager pm, String packageName) {
         try {
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+                // getLongVersionCode contains versionCode in the lower 32bits and versionCodeMajor
+                // in the higher 32 bits. Casting to int will give us the lower 32 bits.
                 return (int) pm.getPackageInfo(packageName, 0).getLongVersionCode();
             }
             return pm.getPackageInfo(packageName, 0).versionCode;
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ChromeLegacyUtils.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ChromeLegacyUtils.java
@@ -14,6 +14,7 @@
 package com.google.androidbrowserhelper.trusted;
 
 import android.content.pm.PackageManager;
+import android.os.Build;
 
 import java.util.Arrays;
 import java.util.List;
@@ -143,8 +144,12 @@ public class ChromeLegacyUtils {
         return getVersionCode(pm, packageName) >= version;
     }
 
+    @SuppressWarnings("deprecation")
     static int getVersionCode(PackageManager pm, String packageName) {
         try {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+                return (int) pm.getPackageInfo(packageName, 0).getLongVersionCode();
+            }
             return pm.getPackageInfo(packageName, 0).versionCode;
         } catch (PackageManager.NameNotFoundException e) {
             return 0;
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/SharedPreferencesTokenStore.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/SharedPreferencesTokenStore.java
@@ -1,0 +1,87 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.androidbrowserhelper.trusted;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+import android.util.Base64;
+
+import androidx.annotation.Nullable;
+import androidx.browser.trusted.Token;
+import androidx.browser.trusted.TokenStore;
+
+/**
+ * Implements a {@link TokenStore} that uses {@link SharedPreferences} as the storage mechanistm
+ * for the {@link Token}.
+ */
+public class SharedPreferencesTokenStore implements TokenStore {
+    private static final String SHARED_PREFERENCES_NAME = "com.google.androidbrowserhelper";
+    private static final String KEY_TOKEN =
+            "com.google.androidbrowserhelper.trusted.SharedPreferencesTokenStore.TOKEN";
+
+    private Context mContext;
+
+    /**
+     * Creates a new SharedPreferencesTokenStore
+     *
+     * @param mContext The {@link Context} where the {@link SharedPreferences} will be stored.
+     */
+    public SharedPreferencesTokenStore(Context mContext) {
+        this.mContext = mContext;
+    }
+
+    /**
+     * This persists the given {@link Token} on a {@link SharedPreferences}.
+     * Subsequent calls will overwrite the previously given {@link Token}.
+     *
+     * @param token The token to persist. It may be {@code null} to clear the storage.
+     */
+    @Override
+    public void store(@Nullable Token token) {
+        SharedPreferences preferences =
+                mContext.getSharedPreferences(SHARED_PREFERENCES_NAME, Context.MODE_PRIVATE);
+
+        // Clear the preference if the token is null
+        if (token == null) {
+            preferences.edit().remove(KEY_TOKEN).apply();
+            return;
+        }
+
+        String encodedToken = Base64.encodeToString(token.serialize(), Base64.DEFAULT);
+        preferences.edit()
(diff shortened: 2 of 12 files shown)


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ChromeLegacyUtils.java
```diff
@@ -143,8 +144,12 @@ private static boolean checkChromeVersion(PackageManager pm, String packageName,
         return getVersionCode(pm, packageName) >= version;
     }
 
+    @SuppressWarnings("deprecation")
     static int getVersionCode(PackageManager pm, String packageName) {
         try {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+                return (int) pm.getPackageInfo(packageName, 0).getLongVersionCode();
```
PEConn: Could you add a comment here saying why this works and that we only guarantee it works with Chrome?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -222,11 +223,15 @@ private void launchWhenSplashScreenReady(TrustedWebActivityIntentBuilder builder
...
-        Intent intent = builder.build(mSession);
+        Intent intent = builder.build(mSession).getIntent();
         ContextCompat.startActivity(mContext, intent, null);
+
         // Remember who we connect to as the package that is allowed to delegate notifications
         // to us.
-        TrustedWebActivityService.setVerifiedProvider(mContext, mProviderPackage);
+        Token token = Token.create(mProviderPackage, mContext.getPackageManager());
+        TokenStore tokenStore = new SharedPreferencesTokenStore(mContext);
+        tokenStore.store(token);
```
PEConn: Could you move these three lines into a static method on TokenStore (so that this file only knows about TokenStore, not Token), like `setVerifiedProvider`?

PEConn: Or if we want to make testing/customizing easier:

1. Create an instance method `setVerifiedProvider` on TokenStore.
2. Pass a TokenStore in to TwaLauncher in the constructor.

(though I'm happy to leave this until we actually need to test things).

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TrustedWebActivityService.java
```diff
@@ -14,11 +14,18 @@
...
 
-import androidx.browser.trusted.TrustedWebActivityService;
+import androidx.annotation.NonNull;
+import androidx.browser.trusted.TokenStore;
 
 /**
- * An extension of {@link TrustedWebActivityService} to allow us to extend its functionality in the
- * future without requiring clients to change their code.
+ * An extension of {@link androidx.browser.trusted.TrustedWebActivityService} to allow us to extend
+ * its functionality in the future without requiring clients to change their code.
```
PEConn: This comment is no longer accurate - this class now provides implementation for the TrustedWebActivityService.

I remember creating this class early on so that we could make this change in a non breaking way. I had hoped that people would include the `DelegationService` in their manifest instead of AndroidX's `TrustedWebActivityService`. So I think that if we just left the name as-is, this wouldn't be a breaking change (or at least the breaking change would be in AndroidX, making TrustedWebActivityService abstract, not in the Android Browser Helper Library).

Can you remind me which part of the TrustedWebActivityService you think didn't fit the name DelegationService?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/SharedPreferencesTokenStore.java
```diff
@@ -0,0 +1,87 @@
...
+import androidx.browser.trusted.Token;
+import androidx.browser.trusted.TokenStore;
+
+/**
+ * Implements a {@link TokenStore} that uses {@link SharedPreferences} as the storage mechanistm
+ * for the {@link Token}.
+ */
+public class SharedPreferencesTokenStore implements TokenStore {
+    private static final String SHARED_PREFERENCES_NAME = "com.google.androidbrowserhelper";
+    private static final String KEY_TOKEN =
```
PEConn: The token is going to be stored in a Preferences file called "com.google.androidbrowserhelper", so I think the first half of `KEY_TOKEN` is redundant.

(But up to you, I'm not too fussed.)

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/SharedPreferencesTokenStore.java
```diff
@@ -0,0 +1,87 @@
...
+            "com.google.androidbrowserhelper.trusted.SharedPreferencesTokenStore.TOKEN";
+
+    private Context mContext;
+
+    /**
+     * Creates a new SharedPreferencesTokenStore
```
PEConn: nit: full stop.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/SharedPreferencesTokenStore.java
```diff
@@ -0,0 +1,87 @@
...
+        SharedPreferences preferences =
+                mContext.getSharedPreferences(SHARED_PREFERENCES_NAME, Context.MODE_PRIVATE);
+
+        // Clear the preference if the token is null
+        if (token == null) {
+            preferences.edit().remove(KEY_TOKEN).apply();
+            return;
+        }
+
+        String encodedToken = Base64.encodeToString(token.serialize(), Base64.DEFAULT);
```
PEConn: In Chrome [we use](https://source.chromium.org/chromium/chromium/src/+/master:chrome/android/java/src/org/chromium/chrome/browser/browserservices/permissiondelegation/TrustedWebActivityPermissionStore.java;l=244?q=trustedwebactivitypermissionstore&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2Fsearch%2F) `NO_WRAP | NO_PADDING`. Do you think it matters?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TrustedWebActivityService.java
andreban: We can keep it as `DelegationService`, but the break will come from replacing the `androidx.browser.trusted.TrustedWebActivityService` with `com.google.androidbrowserhelper.trusted.TrustedWebActivityService`. svgomg-twa and llama-pack use TrustedWebActivityService. :(

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/SharedPreferencesTokenStore.java
andreban: Those seem to be used to save some bytes when encoding the content. It probably has larger impact when encoding things at scale. Adding this has no downsides other than making the string less human readable, which is not a requirement in this case. I'll add it.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
andreban: I'm creating a `setVerifiedProvider` instance method. The bit that is a not as nice is that I have to pass a `SharedPreferencesTokenStore` in the constructor, as the interface doesn't have `setVerifiedProvider`

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ChromeLegacyUtils.java
andreban: From the Javadoc: `Return versionCode and versionCodeMajor combined together as a single long value. The versionCodeMajor is placed in the upper 32 bits.` - It seems this should work with any Android App, as the lower 32 bits are effectively the versionCode?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TrustedWebActivityService.java
andreban: done

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/SharedPreferencesTokenStore.java
andreban: done

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/SharedPreferencesTokenStore.java
andreban: done

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/SharedPreferencesTokenStore.java
andreban: done

andreban:
@PEConn PTAL

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ChromeLegacyUtils.java
PEConn: Ah, fair enough then.

PR #51, opened 2020-01-03
Adds support for Dark Mode

### LINKED ISSUE #50: Add Dark Mode support
[`setColorSchemeParams`][1] and [`setColorScheme`][2] should as expected on TWAs, and allow to customize the navigation bar and URL bar (when navigating off origin).

- This [segment][3] of code sets the colour and can be extended to add the Color Schemes.
- Relevant fields  need to be added to [`LauncherActivityMetadata`][4].

@PEConn, FYI

[1]: https://developer.android.com/reference/androidx/browser/customtabs/CustomTabsIntent.Builder.html#setColorSchemeParams(int,%20androidx.browser.customtabs.CustomTabColorSchemeParams)

[2]: https://developer.android.com/reference/androidx/browser/customtabs/CustomTabsIntent.Builder.html#setColorScheme(int)

[3]: https://github.com/GoogleChrome/android-browser-helper/blob/master/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java#L136-L139

[4]: https://github.com/GoogleChrome/android-browser-helper/blob/master/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivityMetadata.java

andreban:
- Supports selecting a dark mode color for the status bar and the
  navigation bar.

Closes #50

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
@@ -26,6 +26,8 @@ import com.google.androidbrowserhelper.trusted.splashscreens.PwaWrapperSplashScr
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.AppCompatActivity;
+import androidx.browser.customtabs.CustomTabColorSchemeParams;
+import androidx.browser.customtabs.CustomTabsIntent;
 import androidx.browser.trusted.TrustedWebActivityIntentBuilder;
 import androidx.browser.trusted.TrustedWebActivityService;
 import androidx.core.content.ContextCompat;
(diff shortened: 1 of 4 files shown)


khmyznikov:
Can you do the same for SPLASH_SCREEN_BACKGROUND_COLOR?

andreban:
I think this one requires a change beyond this library. @PEConn can confirm.

PEConn:
So while things like navbar theme colours need to be passed over to the browser, the splash screen colour is used to create a splash screen (in the TWA) and then handed over to the browser. Luckily however, the splashscreen is only relevant for a short amount of time so we could just add logic like this in LauncherActivity (around line 131):

```
if (mode = LIGHT) {
  splashScreenColor = lightSplashColor;
} else {
  splashScreenColor = darkSplashColor;
}
```

andreban:
@PEConn is right. We may be able to override the colour using qualified resources on Android. Will check if it's possible and update.

khmyznikov:
So what's about splash screen color?

uwolfer:
I added dark mode to a PWA bundled with bubblewrap, including splash screen. The changes were quite simple - but took some time to get into it. Sharing here the key changes I did - probably it helps somebody else:
1. copied all `res/drawable` folders to `res/drawable-night-*`, and adjusted `splash.png`
2. added `res/values-night/colors.xml` which defined the `backgroundColor`: `<color name="backgroundColor">#000000</color>`
3. adjusted my `LauncherActivity.java` with an ugly workaround because of a [bug in android-browser-helper](https://github.com/GoogleChrome/android-browser-helper/issues/461) as follows:

```java
+import android.content.SharedPreferences;
 import android.content.pm.ActivityInfo;
 import android.net.Uri;
 import android.os.Build;
@@ -30,6 +31,10 @@
 
     @Override
+        // workaround for this bug: https://github.com/GoogleChrome/android-browser-helper/issues/461
+        SharedPreferences splashImagePrefs = this.getApplicationContext().getApplicationContext().getSharedPreferences("splashImagePrefs", 0);
+        splashImagePrefs.edit().remove("lastUpdateTime").apply();
+
```

khmyznikov:
@uwolfer this is so cool! Do you considered to contribute into bubblewrap with this fix?

uwolfer:
@khmyznikov IMO my change is not really a fix, but more an ugly workaround, which prevents caching of the splash screen. I would like to get this properly fixed in android-browser-helper: https://github.com/GoogleChrome/android-browser-helper/issues/461
I am quite sure where the underlying issue is, and tried to describe it. If you have any connections at Chrome / android-browser-helper, it would be nice if you could get them to look into the issue. Once that is fixed, we could think about improving dark mode in bubblewrap, as there are really only a few changes required, and most of the work is generating proper splash images.

khmyznikov:
@uwolfer how bad is caching prevention for splash? Is it leading to what?

PR #53, opened 2020-01-13
Provide an Intent to the browser to focus the TWA.

PEConn:
Hey,

This is the long awaited counterpart to https://chromium-review.googlesource.com/c/chromium/src/+/1849682

(was https://github.com/GoogleChrome/android-browser-helper/pull/41 before I failed rebasing.)


andreban:
@PEConn I guess the best practice would be to add this by default to apps? (If so, we should probably create an issue to update the llama-pack template, once this makes into a new version)

PEConn:
Yeah, that would be helpful (after a bit of further testing).

PR #54, opened 2020-01-14
Create utilities to call PlayBilling querySkuDetailsAsync in another process

PEConn:
Hey Andre,

I have this working locally with a test Play Billing set up. My next step is to add a more involved Android test that makes Play Billing calls.


build.gradle
```diff
@@ -16,18 +16,20 @@
...
+
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:3.4.2'
-        
+
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
-    }
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version" }
```
andreban: I'm assuming there's not way to add the Kotlin stuff only to the playbilling module (or maybe we want to use Kotlin in other places as well?)

PEConn: After I've got the library in some rough shape I'm going to bundle it, include it in Chrome and see what the binary size impact is. If it's not too big, then I'm happy spreading Kotlin to other parts of ABH, otherwise I'll probably have to rewrite this all in Java and we can get rid of Kotlin.

OTOH, maybe we could just spread Kotlin everywhere and only revert the play billing part if needed.

andreban: SGTM

build.gradle
```diff
@@ -16,18 +16,20 @@
...
 
-buildscript {
+buildscript { 
+    ext.kotlin_version = '1.3.50'
+
     repositories {
         google()
         jcenter()
-        
+
```
rknoll: nit: let's clean up this file a bit :-)

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/bundles/QuerySkuDetailsCall.kt
```diff
@@ -0,0 +1,31 @@
...
+        val params: SkuDetailsParams,
+        val listener: SkuDetailsResponseListener) {
+
+    fun toBundle() = makeBundle(CLASS_NAME) {
+        putBundle(PARAMS_KEY, params.toBundle())
+        putBinder(CALLBACK_KEY, listener.toBinder())
+    }
+
+    companion object {
+        fun fromBundle(bundle: Bundle): QuerySkuDetailsCall {
```
rknoll: If you want to be consistent with the other files in /bundles/ you could put this on the Bundle too:
`fun Bundle.toQuerySkuDetailsCall(): QuerySkuDetailsCall { ... }`

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/bundles/SkuDetailsResponseListener.kt
```diff
@@ -0,0 +1,44 @@
...
+fun binderToListener(binder: IBinder): SkuDetailsResponseListener {
+    val callback = ICallback.Stub.asInterface(binder)
+
+    return SkuDetailsResponseListener { billingResult, skuDetails ->
+        val argsBundle = makeBundle(CLASS_NAME) {
+            put(BILLING_RESULT_KEY, billingResult.toBundle())
+            put(SKU_DETAILS_LIST, skuDetails!!.toParcelableArray())
+        }
+
+        Log.d("Peter", "Binder: $binder, Callback: $callback")
```
rknoll: Remove?

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/bundles/QuerySkuDetailsCall.kt
PEConn: I'd left this one as-is at the moment because QuerySkuDetailsCall is going to be the single public class and probably called from Java and `QuerySkuDetailsCall.fromBundle(Bundle)` looks nicer than `QuerySkuDetailsCall.toQuerySkuDetailsCall(Bundle)`.

The API is fairly alpha at the moment though.

PEConn: (Got this working once I added the missing `@JvmStatic` annotation.)

PR #57, opened 2020-01-17
Fixes asset_staments on twa-multi-domain

andreban:
- strings.xml using common.share_location instead of
  common.handle_all_urls


pianomister:
`share_location` is also used in the documentation here:
https://github.com/GoogleChrome/android-browser-helper/tree/master/demos/twa-multi-domain

andreban:
> https://github.com/GoogleChrome/android-browser-helper/blob/master/demos/twa-multi-domain/README.md

Fixed that too

PR #67, opened 2020-02-14
WebView Fallback demo handles custom colors

andreban:
- Handles navigation bar color, status bar color and uses the
  status bar color when launching CCT.
- Still needs changes to support dark mode

### DIFF
--- a/demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/MyLauncherActivity.java
+++ b/demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/MyLauncherActivity.java
@@ -16,8 +16,6 @@ package com.google.browser.examples.twawebviewfallback;
 
 import android.content.Intent;
 
-import androidx.core.content.ContextCompat;
-
 import com.google.androidbrowserhelper.trusted.LauncherActivity;
 import com.google.androidbrowserhelper.trusted.LauncherActivityMetadata;
 import com.google.androidbrowserhelper.trusted.TwaLauncher;
@@ -27,15 +25,8 @@ public class MyLauncherActivity extends LauncherActivity {
     @Override
     protected TwaLauncher.FallbackStrategy getFallbackStrategy() {
         return (context, twaIntentBuilder, mProviderPackage, completionCallback) -> {
-            Intent intent = new Intent(context, WebViewFallbackActivity.class);
-            intent.putExtra(WebViewFallbackActivity.KEY_LAUNCH_URI, twaIntentBuilder.getUri());
-
-            LauncherActivityMetadata metadata = LauncherActivityMetadata.parse(this);
-            intent.putExtra(WebViewFallbackActivity.STATUS_BAR_COLOR,
-                    ContextCompat.getColor(this, metadata.statusBarColorId));
-            intent.putExtra(WebViewFallbackActivity.NAVIGATION_BAR_COLOR,
-                    ContextCompat.getColor(this, metadata.navigationBarColorId));
-
+            Intent intent = WebViewFallbackActivity.createLaunchIntent(
+                    this, twaIntentBuilder.getUri(), LauncherActivityMetadata.parse(this));
             startActivity(intent);
             completionCallback.run();
         };
--- a/demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/MyLauncherActivity.java
+++ b/demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/MyLauncherActivity.java
@@ -16,7 +16,10 @@ package com.google.browser.examples.twawebviewfallback;
 
 import android.content.Intent;
 
+import androidx.core.content.ContextCompat;
+
 import com.google.androidbrowserhelper.trusted.LauncherActivity;
+import com.google.androidbrowserhelper.trusted.LauncherActivityMetadata;
 import com.google.androidbrowserhelper.trusted.TwaLauncher;
 
 public class MyLauncherActivity extends LauncherActivity {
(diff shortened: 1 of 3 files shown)


demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/MyLauncherActivity.java
```diff
@@ -26,6 +29,13 @@
         return (context, twaIntentBuilder, mProviderPackage, completionCallback) -> {
             Intent intent = new Intent(context, WebViewFallbackActivity.class);
             intent.putExtra(WebViewFallbackActivity.KEY_LAUNCH_URI, twaIntentBuilder.getUri());
+
+            LauncherActivityMetadata metadata = LauncherActivityMetadata.parse(this);
+            intent.putExtra(WebViewFallbackActivity.STATUS_BAR_COLOR,
+                    ContextCompat.getColor(this, metadata.statusBarColorId));
+            intent.putExtra(WebViewFallbackActivity.NAVIGATION_BAR_COLOR,
+                    ContextCompat.getColor(this, metadata.navigationBarColorId));
```
PEConn: What I'd probably do is create a static method called `createLaunchIntent` in WebViewFallbackActivity that takes a URL and the colours (or maybe the URL and the metadata), so this method would become:

    Intent intent = WebViewFallbackActivity.createLaunchIntent(context,
        twaIntentBuilder.getUri(), LauncherActivityMetadata.parse(this));
    startActivity(intent);
    completionCallback.run();

This way, `KEY_LAUNCH_URI`, `NAVIGATION_BAR_COLOR` and `STATUS_BAR_COLOR` can all be private and only WebViewFallbackActivity needs to know exactly how the Intent is formatted.

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
```diff
@@ -30,15 +31,30 @@
 public class WebViewFallbackActivity extends AppCompatActivity {
     public static final String KEY_LAUNCH_URI =
             "com.google.browser.examples.twawebviewfallback.WebViewFallbackActivity.KEY_LAUNCH_URL";
+    public static final String NAVIGATION_BAR_COLOR =
+            "com.google.browser.examples.twawebviewfallback.WebViewFallbackActivity" +
```
PEConn: Since you're concatenating strings anyway, why not do:

    private static final String KEY_PREFIX =
        "com.google.browser.examples.twawebviewfallback.WebViewFallbackActivity.";
    private static final String KEY_LAUNCH_URI = KEY_PREFIX + "KEY_LAUNCH";
    private static final String KEY_NAVIGATION_BAR_COLOR = KEY_PREFIX + "NAVIGATION_BAR_COLOR";

Note, also:

* I made them private according to my comment in `MyLauncherActivity`.
* `NAVIGATION_BAR_COLOR` gets `KEY_` as a prefix (like `KEY_LAUNCH_URI`).
* The *value* of `KEY_LAUNCH_URI` doesn't contain `KEY_` (since that seems a bit odd).

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
```diff
@@ -30,15 +31,30 @@
...
+                    ".NAVIGATION_BAR_COLOR";
+    public static final String STATUS_BAR_COLOR =
+            "com.google.browser.examples.twawebviewfallback.WebViewFallbackActivity" +
+                    ".STATUS_BAR_COLOR";
+
 
     private Uri mLaunchUrl;
+    private int mStatusBarColor;
 
+    @SuppressLint("SetJavaScriptEnabled")
```
PEConn: What's this about?

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
```diff
@@ -30,15 +31,30 @@
...
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
         this.mLaunchUrl = this.getIntent().getParcelableExtra(KEY_LAUNCH_URI);
 
+        int navigationBarColor = this.getIntent().getIntExtra(NAVIGATION_BAR_COLOR, 0);
+        getWindow().setNavigationBarColor(navigationBarColor);
```
PEConn: Does this cause a behaviour change when an app doesn't have the colours set?

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
andreban: Android Studio linter alerts when we do `WebViewSettings#setJavaScritpEnabled(true)`. But this is required in this case.

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/MyLauncherActivity.java
andreban: done

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
andreban: Changed to only set colors if they exist in the Intent

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
andreban: done

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
PEConn: But that still would cause a behaviour change because there's always going to be the extras in the Intent. (We put the defaults from LauncherActivityMetadata in there.)

This isn't necessarily a bad thing though - maybe we want to do it.

andreban: So, it seems we'd be replicating the behaviour that have today? I think this makes sense. The alternative would be updating `LauncherActivityMetadata` to get its default values from the system. If we did, the demo would automatically get the new value.

PEConn: Fair enough, that's fine by me.

PR #68, opened 2020-02-17
Custom header ergonomics

andreban:
- Another (maybe better?) approach would be adding support for
this in the TwaIntentBuilder, which is part of AndroidX.
- We should also explore adding a specific approach vs a callback to modify the underlying intent, but I wonder which other modifications would be desirable.

@PEConn WDYT?

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
@@ -32,6 +32,8 @@ import androidx.browser.trusted.TrustedWebActivityIntentBuilder;
 import androidx.browser.trusted.TrustedWebActivityService;
 import androidx.core.content.ContextCompat;
 
+import java.util.Map;
+
 /**
  * A convenience class to make using Trusted Web Activities easier. You can extend this class for
  * basic modifications to the behaviour.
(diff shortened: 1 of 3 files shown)


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -41,19 +46,41 @@
...
     };
 
+    private static void applyCustomHeaders(
+            Intent intent, @Nullable Map<String, String> customHeaders) {
+        if (customHeaders == null) {
+            return;
+        }
+
+        Bundle headerBundle = new Bundle();
+        for (Map.Entry<String, String> header: customHeaders.entrySet()) {
```
PEConn: You can also write this as:

    for (String key: customHeaders.keySet()) {
        String value = customHeaders.get(key);

Up to you though.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -156,7 +158,8 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
         mTwaLauncher.launch(twaBuilder,
                 mSplashScreenStrategy,
                 () -> mBrowserWasLaunched = true,
-                getFallbackStrategy());
+                getFallbackStrategy(),
+                getCustomHeaders());
```
dalexanco, line 162: since the goal of these headers is to be attached to the intent, shouldn't be part of the intent builder ?
We could extends TrustedWebActivityIntentBuilder and add properties to it...

andreban: `TrustedWebActivityIntentBuilder` is part of `androidx.browser` and we don't have access to modify it here. It would be nice to have it there, but I'd rather not be blocked by a change upstream.

andreban:
Those changes are incompatible with how the API will work in the future. Closing the PR.

PR #76, opened 2020-03-08
update gradle and avoid leaked service connection in sample

khatv911:
- Refactor sample code to avoid leaking service connection
- version updated for gradle wrapper and build tool

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ buildscript {
 
     }
     dependencies {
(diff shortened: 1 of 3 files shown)


khatv911:
@googlebot I signed it!

PEConn:
In general I'd prefer code changes and bumping the gradle version to be kept separate (so that if bumping the gradle version breaks something, we can revert it easily), but this CL isn't that complex so I'm not too fussed.

Code LGTM too.

Issue #15, opened 2019-09-04
WebView fallback


PEConn:
Investigate providing a WebView fallback when the device does not have a TWA provider.

* Have a look at current usage.
* Consider APK size increase.
* Consider WebView Web API support.

andreban:
On the normal flow, the browser will validate Digital Asset Links. In this case, everything is happening in the same application. Should the Support Library perform the validation in this case?

willyau:
This enhancement will be a great improvement for users. The current implementation can fallback to the user's browser. This causes a very jarring and disruptive experience. The content is loaded in the browser. Often times, this has many tabs opened. And switching back to the app (from app switcher) doesn't do anything (just shows homescreen).

PEConn:
Yeah, that makes sense - a demo people can copy and give feedback on, then
potentially integrating it into the library.

andreban:
#46 kicked off a demo implementation. It's a naive implementation for now and more work is needed for it to be a good fallback implementation.

naveedahmed1:
This would really help in improving overall user experience with TWA apps.

strygo:
The WebView fallback sample is a good starting point, but there are aspects of it that need further functionality to bring the WebView experience closer to parity with TWAs:

- Handle orientation changes so that page state isn't lost
- Handle onPause/onResume
- Handle onBackPressed
- Handle onSaveInstanceState
- Enable domstorage, database, etc.
- The navigation bar color behavior doesn't seem right. As of now, the sample uses transparent, which is optimal for the splash screen but not what you'd expect when the WebView is shown

andreban:
Yes, that's why we haven't added it as a default to the library yet. We haven't been able to come back to this yet. Any contributions to improve the WebView fallback demo will be greatly appreciated, as we intend to merge it into the Support Library.

strygo:
OK. I am also experiencing an issue with audio playback that I haven't resolved. These are showing in the debug logs:

W/AudioManager: Use of stream types is deprecated for operations other than volume control
W/AudioManager: See the documentation of requestAudioFocus() for what to use instead with android.media.AudioAttributes to qualify your playback use case

andreban:
What does the issue you are experience looks like?

One thing to keep in mind is that the WebView fallback will probably work on a best-effort to support PWA features - The WebView doesn't support all Web Platform APIs and, in many cases, feature detection also doesn't work as expected.

[...]

PR #77, opened 2020-03-09
Improves the WebView Fallback Demo

andreban:
- Implements onStop() and onStart()
- Implements onSaveInstanceBundle()
- Handles multiple origins
- Enables domStorage and database

### DIFF
--- a/demos/twa-webview-fallback/src/main/AndroidManifest.xml
+++ b/demos/twa-webview-fallback/src/main/AndroidManifest.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.google.browser.examples.twawebviewfallback">
+    <uses-permission android:name="android.permission.INTERNET"/>s
 
     <application
         android:allowBackup="true"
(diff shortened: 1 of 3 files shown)


andreban:
Improves on #15

demos/twa-webview-fallback/src/main/java/com/google/browser/examples/twawebviewfallback/WebViewFallbackActivity.java
```diff
@@ -139,6 +199,28 @@ public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request
...
+            private boolean matchExtraOrigins(Uri navigationUri) {
+                for (Uri uri: mExtraOrigins) {
+                    if (uriOriginsMatch(uri, navigationUri)) {
+                        return true;
+                    }
+                }
+                return false;
+            }
+
+            private boolean uriOriginsMatch(Uri uriA, Uri uriB) {
```
rayankans: I think you should also check that the ports match. And maybe instead of comparing the schemes, make sure they are both https?

andreban: Added the check for `port`. I'm now checking for `https` in `launchUrl` when starting the Activity and ignoring additionalOrigins that are not `https`. Kept `getScheme().equals()` here.

demos/twa-webview-fallback/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.google.browser.examples.twawebviewfallback">
+    <uses-permission android:name="android.permission.INTERNET"/>s
```
strygo: "s" typo

andreban: fixed

PR #78, opened 2020-03-09
Moves the WebViewFallback into the library

andreban:
- Moves the WebViewFallbackActivity into the Library
- Prepares the default LauncherActivity to accept a meta-tag
  with the fallback to be used.
- When the fallabck is "webview", launches the WebView fallback.
  Otherwise, launches a custom tab.


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/WebViewFallbackActivity.java
```diff
@@ -84,17 +85,25 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
-        if (getIntent().hasExtra(KEY_NAVIGATION_BAR_COLOR)) {
-            int navigationBarColor = this.getIntent().getIntExtra(
-                    KEY_NAVIGATION_BAR_COLOR, 0);
-            getWindow().setNavigationBarColor(navigationBarColor);
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            if (getIntent().hasExtra(KEY_NAVIGATION_BAR_COLOR)) {
+                int navigationBarColor = this.getIntent().getIntExtra(
+                        KEY_NAVIGATION_BAR_COLOR, 0);
+                    getWindow().setNavigationBarColor(navigationBarColor);
```
rayankans: unindent

demos/twa-webview-fallback/src/main/AndroidManifest.xml
```diff
@@ -41,6 +41,9 @@
             <meta-data android:name="android.support.customtabs.trusted.ADDITIONAL_TRUSTED_ORIGINS"
                 android:resource="@array/additional_trusted_origins" />
 
+            <meta-data android:name="android.support.customtabs.trusted.FALLBACK_STRATEGY"
+                android:value="webview" />
```
rayankans: is there a reason why we are preferring webviews over customtabs?

I'm not sure what the compat is like, but presumably customtabs should be preferred due to the matching API parity

andreban:
Moves forward #15

demos/twa-webview-fallback/src/main/AndroidManifest.xml
andreban: We're preferring CustomTabs over WebView. The demo shows how to change the default, so it has the meta-tag.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/WebViewFallbackActivity.java
andreban: fixed

demos/twa-webview-fallback/src/main/AndroidManifest.xml
andreban: TL;DR is that yes, compat is bad and developers shouldn't expect the WebView fallback to work seamlessly, but some users get frustrated when they get the Custom Tabs experience, so developers having been asking for a WebView fallback. We still recommend CustomTabs for compatibility. More browsers are working on Trusted Web Activity support, so the need for a fallback should decrease soon.

demos/twa-webview-fallback/src/main/res/values/strings.xml
```diff
@@ -19,12 +19,25 @@
...
             \"relation\": [\"delegate_permission/common.handle_all_urls\"],
             \"target\": {
                 \"namespace\": \"web\",
                 \"site\": \"https://pwa-directory.appspot.com/\"}
+        }],
+        [{
+            \"relation\": [\"delegate_permission/common.handle_all_urls\"],
+            \"target\": {
+                \"namespace\": \"web\",
+                \"site\": \"https://pm5-monitor-c63a2.firebaseapp.com/\"}
```
PEConn: What's this about?

demos/twa-webview-fallback/src/main/AndroidManifest.xml
PEConn: How come we're using webview instead of WebView? (Or is it just case insensitive?)

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -259,6 +261,16 @@ protected Uri getLaunchingUrl() {
     }
 
     protected TwaLauncher.FallbackStrategy getFallbackStrategy() {
+        if (FALLBACK_TYPE_WEBVIEW.equalsIgnoreCase(mMetadata.fallbackStrategyType)) {
```
PEConn: Could we make this a member of TwaLauncher like CCT_FALLBACK_STRATEGY (eg, WEBVIEW_FALLBACK_STRATEGY)?

demos/twa-webview-fallback/src/main/AndroidManifest.xml
andreban: The value is defined inside LauncherActivity:

```
    private static final String FALLBACK_TYPE_WEBVIEW = "webview";
```

demos/twa-webview-fallback/src/main/res/values/strings.xml
andreban: The second is a PWA inside PWA Directory to make testing navigation from one domain to the other easy.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
andreban: done

PR #81, opened 2020-03-31
Verify the caller of PaymentActivity

PEConn:
Hey Andre and Rouslan,

Please take a look at this WIP CL. At the moment it implements `IS_READY_TO_CALL`, which allows Chrome to check whether the TWA is going to accept payment requests from it.

The second part is checking which app launches PaymentActivity and finish immediately if it's not verified (it's only a few lines of code, but unfortunately you can't launch an Activity with `startActivityForResult` from an `ActivityTestRule` so I'm going to have to redo some tests).

(I also need to add the newly created Service to the AndroidManifest.xml of the demo.)

### DIFF
--- a/playbilling/src/androidTest/java/com/google/androidbrowserhelper/playbilling/provider/WrapperActivity.java
+++ b/playbilling/src/androidTest/java/com/google/androidbrowserhelper/playbilling/provider/WrapperActivity.java
@@ -1,0 +1,67 @@
+package com.google.androidbrowserhelper.playbilling.provider;
+
+import android.content.Intent;
+import android.os.Bundle;
+import android.os.Handler;
+import android.util.Log;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import androidx.annotation.Nullable;
+import androidx.appcompat.app.AppCompatActivity;
+
+/**
+ * An Activity that launches another Activity with {@link #startActivityForResult}.
+ *
+ * Originally I'd attempted to have this Activity launched by an ActivityTestRule, but that seemed
+ * unable to handle an Activity that immediately launched another Activity.
+ *
+ * I agree that having statics all over the place is ugly, but it's test code and it works.
+ */
+public class WrapperActivity extends AppCompatActivity {
+    // Can't use 0, 1 or -1 as they're valid values.
+    private static final int INVALID_RESULT_CODE = 666;
+    private static int sResultCode = INVALID_RESULT_CODE;
+
+    private static CountDownLatch sLaunchLatch = new CountDownLatch(1);
+    private static CountDownLatch sFinishLatch = new CountDownLatch(1);
+
+    public static final String EXTRA_INNER_INTENT = "inner_intent";
+
+    @Override
+    public void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        sResultCode = INVALID_RESULT_CODE;
+
+        Intent innerIntent = getIntent().getParcelableExtra(EXTRA_INNER_INTENT);
+        startActivityForResult(innerIntent, 0);
+
+        sLaunchLatch.countDown();
+    }
+
+    @Override
+    protected void onActivityResult(int requestCode, int resultCode, @Nullable Intent data) {
+        sResultCode = resultCode;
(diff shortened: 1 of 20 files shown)


PEConn:
I've now added verifying the caller to the Activity as well.

playbilling/src/androidTest/java/com/google/androidbrowserhelper/playbilling/provider/WrapperActivity.java
```diff
@@ -0,0 +1,67 @@
...
+import androidx.annotation.Nullable;
+import androidx.appcompat.app.AppCompatActivity;
+
+/**
+ * An Activity that launches another Activity with {@link #startActivityForResult}.
+ *
+ * Originally I'd attempted to have this Activity launched by an ActivityTestRule, but that seemed
+ * unable to handle an Activity that immediately launched another Activity.
+ *
+ * I agree that having statics all over the place is ugly, but it's test code and it works.
```
rsolomakhin: Since it's test code, should the name of the class reflect that somehow?

playbilling/src/main/aidl/com/google/androidbrowserhelper/playbilling/IIsReadyToPayService.aidl
```diff
@@ -0,0 +1,7 @@
+package com.google.androidbrowserhelper.playbilling;
```
rsolomakhin: Does the package name here have to match the AIDL file in Chromium repo?

playbilling/src/main/aidl/com/google/androidbrowserhelper/playbilling/IIsReadyToPayServiceCallback.aidl
```diff
@@ -0,0 +1,5 @@
+package com.google.androidbrowserhelper.playbilling;
```
rsolomakhin: Ditto

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/PaymentVerifier.java
```diff
@@ -0,0 +1,45 @@
...
+     * A package can only trigger payment requests if it is the verified provider for the Trusted
+     * Web Activity.
+     */
+    static boolean shouldAllowPayments(@NonNull Context context, @Nullable String packageName,
+            @NonNull String logTag) {
+        // TODO: Should I also check whether the TWA is currently running? If so, how?
+        if (packageName == null) return false;
+
+        TokenStore tokenStore = new SharedPreferencesTokenStore(context);
+        Token verifiedPackage = tokenStore.load();
```
rsolomakhin: Neato!

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/MockBillingWrapper.java
```diff
@@ -21,21 +23,25 @@
 
     @Override
     public void connect() {
+        Log.d("Peter", "a");
```
andreban: We probably want to remove those.

playbilling/src/androidTest/java/com/google/androidbrowserhelper/playbilling/provider/WrapperActivity.java
PEConn: It's in the androidTest build target - it seems a bit odd to have everything in there include the word Test in its name.

playbilling/src/main/aidl/com/google/androidbrowserhelper/playbilling/IIsReadyToPayService.aidl
PEConn: I'm not actually sure - I don't fully understand AIDL matching. I'll change it just to be safe. (I'd run the tests I'd created, but not manually tested with Chrome yet.)

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/MockBillingWrapper.java
PEConn: :man_facepalming: 

I promise I do know how to use a debugger - I was just trying to figure out which parts of a test were taking so long.

PR #83, opened 2020-04-06
Support for Web Share Target

flunderpero:
I cherry picked psmakov's commit (223a17c) and adapted the code.

Everything works ok on most devices, but there are still problems where things go wrong. Perhaps someone with more knowledge about Android, androidx or android-browser-helper could help me out here.

*Problem 1*: On my Pixel 4XL w/ Android 10 (QQ2A.200305.003) and Google Chrome (80.0.3987.162) and my Nexus 10 tablet w/ Android 8 (8.1.0 Pixel C Build OPM8.190605.005) and the same Google Chrome version sharing with our TWA _sometimes_ works for images (only) and only if our TWA is already running. (On a Pixel 3 XL w/ Android 10 (QQ2A.200305.002) and the same Google Chrome version everything works as expected.)
What happens is that I get a blank splash-screen. When I attach the Google Chrome  device inspector on it, I see that there is only an empty tab open on the device and no service-worker. So clearly the tab has not been started correctly.
In the logs of the phone (via logcat in Android Studio) I just see:
```
2020-04-05 22:04:22.877 1345-1953/? I/ActivityTaskManager: START u0 {act=android.intent.action.SEND typ=image/jpeg flg=0x10000001 cmp=com.clingmobileapp/.ClingApplicationLauncherShare clip={image/jpeg U:content://com.google.android.apps.photos.contentprovider/0/1/content%3A%2F%2Fmedia%2Fexternal%2Fimages%2Fmedia%2F3537/REQUIRE_ORIGINAL/NONE/image%2Fjpeg/190640420} (has extras)} from uid 10173
```
Everything is working fine if I disable Asset Link checking as described [here](https://github.com/GoogleChromeLabs/svgomg-twa#enabling-debug). I even tried to disable the splash screen, but still it does not work on my phone. It works with a Google Pixel 4 / Android 10 emulator btw.

*Problem 2*: This is more of a UX problem. If the installed Google Chrome version does not support Share API v2 w/ TWA I would like to disable the share-intent or at least be able to react to it and show an info message to the user that an update of Google Chrome is necessary to use sharing. Is there any way to know, which Chrome version is available?

### DIFF
--- a/androidbrowserhelper/src/androidTest/AndroidManifest.xml
+++ b/androidbrowserhelper/src/androidTest/AndroidManifest.xml
@@ -76,13 +76,20 @@
                   android:label="Test Launcher Activity"
                   android:theme="@style/Theme.AppCompat.Light"
                   android:enabled="false">
-
             <meta-data android:name="android.support.customtabs.trusted.DEFAULT_URL"
                        android:value="https://www.test.com/default_url/" />
-
             <meta-data
                 android:name="android.support.customtabs.trusted.STATUS_BAR_COLOR"
                 android:resource="@color/status_bar_color" />
+            <meta-data
+                android:name="android.support.customtabs.trusted.METADATA_SHARE_TARGET"
+                android:resource="@string/share_target"/>
+            <intent-filter>
+                <action android:name="android.intent.action.SEND" />
+                <action android:name="android.intent.action.SEND_MULTIPLE" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data android:mimeType="image/*" />
+            </intent-filter>
         </activity>
 
         <!-- For TWA splash screens -->
--- a/androidbrowserhelper/src/androidTest/java/com/google/androidbrowserhelper/trusted/LauncherActivityTest.java
+++ b/androidbrowserhelper/src/androidTest/java/com/google/androidbrowserhelper/trusted/LauncherActivityTest.java
@@ -20,6 +20,7 @@ import static org.junit.Assert.assertFalse;
 import static androidx.browser.customtabs.TrustedWebUtils.EXTRA_LAUNCH_AS_TRUSTED_WEB_ACTIVITY;
 
 import android.content.Context;
+import android.content.Intent;
 import android.net.Uri;
 
 import com.google.androidbrowserhelper.test.R;
@@ -35,7 +36,15 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
 import androidx.browser.customtabs.CustomTabsIntent;
+import androidx.browser.trusted.TrustedWebActivityIntentBuilder;
+import androidx.browser.trusted.sharing.ShareData;
+import androidx.browser.trusted.sharing.ShareTarget;
 import androidx.test.InstrumentationRegistry;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 import androidx.test.filters.MediumTest;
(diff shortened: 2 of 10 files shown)


flunderpero:
@googlebot I signed it!

PEConn:
I had a skim over the code, nothing stands out to me, but I'll need to have a more in depth look through it later - did you end up changing a lot or is it mostly unchanged from the commit you cherry picked?

I don't have any ideas off the top of my head for the first issue - I'll have to look into it a bit further (time permitting).

For your second question, yeah, you can figure out the version of Chrome that is available (look at ChromeLegacyUtils.java). However, I'd prefer we do things a slightly different way.

When Chrome added support for this, we added the [CATEGORY_WEB_SHARE_TARGET_V2][1] to its AndroidManifest. We can then check whether this category is present (we do something similar [here][2]). That way we can handle other browsers supporting this.

Once we find out whether the browser supports it, it should be easy to show a Toast to the users. We could even disable the intent filter for next time ([eg][3]).

[1]: https://developer.android.com/reference/androidx/browser/customtabs/CustomTabsService#CATEGORY_WEB_SHARE_TARGET_V2
[2]: https://github.com/GoogleChrome/android-browser-helper/blob/5f6d124afa28b1693bc2f047f570656a06d057a6/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/splashscreens/SystemBarColorPredictor.java#L111-L120
[3]: https://stackoverflow.com/a/40505582

flunderpero:
@PEConn thanks for your reply.

I did change very little. I mainly fixed a bug in SharingUtils, where we tried to read `intent.getParcelableArrayListExtra(Intent.EXTRA_STREAM)`, which throws an error if there is no streaming data.

I structured the commits in a way that should make it clear, what was changed and what came straight from the commit I cherry picked.

flunderpero:
I just rebased this PR on top of the upstream master. The problems (App won't start sometimes when used as a share target) persist unfortunately. It would be great if someone could step in.

flunderpero:
Might [this](https://github.com/GoogleChrome/android-browser-helper/pull/83/commits/995f939a6f6e6760834553f9950ea45b35ed3346) be the problem?

I had to adapt the way the PWA is launched, otherwise sharing would not work at all. I got this piece of code from the original commit by pshmakov.

andreban:
@flunderpero I'm not sure this is the problem. [`TrustedWebActivityIntent.launchTrustedWebActivity`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-master-dev:browser/browser/src/main/java/androidx/browser/trusted/TrustedWebActivityIntent.java;l=46-49?q=TrustedWebActivityintent) does have have [additional behaviour that seems to handle permission for the Share Target URIs](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-master-dev:browser/browser/src/main/java/androidx/browser/trusted/TrustedWebActivityIntent.java;l=51-56?q=TrustedWebActivityintent). 

Do you have a test app you've been using? I'd like to give it a try myself.

flunderpero:
Since the problem only occurs if I enable Digital Asset Linking, I cannot provide a simple demo-app. (To be precise: If I *disable* the verification in Chrome (`--disable-digital-asset-link-verification-for-url=https://cling.com/c`) everything works fine so far.)

You can join our internal test channel and try the app for yourself. You'll need to sign-up with our service (cling.com), which is a free platform focused on content creation, curation and collaboration.
https://play.google.com/apps/internaltest/4698063214496940258

Then try to share an image, a link or text with the app. It _might_ work if the app is open in the background, but it never works if the app is not running AFAICS. I am currently using a Pixel 4 XL with Android 10 (see above).

You can then compare this with the PWA, which works smoothly all the time.

andreban:
Sounds like the testing is invite-only: `Your account hasn't yet been invited to participate in this app's internal testing program.`

flunderpero:
Sorry, I didn't check that. Now you should be able to download the test version.

mvano:
Hi Peter, I'm looking into this issue. Does it still reproduce? Can I test it with your app? Thanks!

flunderpero:
Hi @mvano, yes it still reproduces (unfortunately). And also yes, you can use our app: https://play.google.com/apps/internaltest/4698063214496940258

mvano:
Thanks @flunderpero! Could you put me on the allowlist for the test app? I signed up with timmytester1337 at gmail.com.

flunderpero:
@mvano I just added you. Please let me know if I can help out in any way.

mvano:
Thanks @flunderpero though I still don't seem to have access. In the meantime, I've landed #160 which has two test sites, one for text and one for photos. Perhaps we can continue on that more accessible basis, if the bug still happens there.

flunderpero:
@mvano That's great news. I updated our app with the lastest main build of android-browser-helper and the problems persist unfortunately. I granted access to timmytester...@ in the console (sorry, I made a mistake with that before). Perhaps you can give it a try and see something more. 

I will take a deeper look into the demos in order to find differences in the usage of the library.

flunderpero:
I adapted my manifest to match the manifest in the demo (I had an activity-alias for example), but the problems are still there.

OTOH, I cannot share an image with the demo-app at all. I see the launcher or a white page and nothing ever happens.

PR #95, opened 2020-05-06
Allow setting Display Mode on AndroidManifest

andreban:
Haven't tested on proper devices yet. Do not merge.

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
@@ -17,6 +17,7 @@ package com.google.androidbrowserhelper.trusted;
 import android.content.Intent;
 import android.graphics.Matrix;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Bundle;
 import android.util.Log;
 import android.widget.ImageView;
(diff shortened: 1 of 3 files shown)


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -267,6 +272,26 @@ protected Uri getLaunchingUrl() {
         return TwaLauncher.CCT_FALLBACK_STRATEGY;
     }
 
+    protected TrustedWebActivityDisplayMode getDisplayMode() {
```
PEConn, line 275: I think this method should be in `LauncherActivityMetadata`. That class should do the parsing from Strings to domain types.

andreban: Cutout mode can have 3 different values, while sticky has 2. That leads to 6 different possible values.  We can either cover all values as strings or specify what be believe are the most common options and let developers override if needed.

andreban: changed.

andreban:
Updated and tested on a proper device. @PEConn PTAL

PR #104, opened 2020-05-28
TwaProviderPicker ignores non-browsers

andreban:
- Some applications seem to declare handling Intents for http://
- Changed the browser detection as seen on https://cs.android.com/android/platform/superproject/+/master:packages/apps/PermissionController/src/com/android/packageinstaller/role/model/BrowserRoleBehavior.java
- Added tests for those cases

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaProviderPicker.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaProviderPicker.java
@@ -101,12 +101,12 @@ public class TwaProviderPicker {
      * supports.
      */
     public static Action pickProvider(PackageManager pm) {
-        // TODO(peconn): Should we use "https://" instead?
+        // Setting the Intent Data as seen at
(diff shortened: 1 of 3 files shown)


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaProviderPicker.java
```diff
@@ -101,12 +101,12 @@ public Action(@LaunchMode int launchMode, @Nullable String provider) {
      * supports.
      */
     public static Action pickProvider(PackageManager pm) {
-        // TODO(peconn): Should we use "https://" instead?
+        // Setting the Intent Data as seen at
+        // https://cs.android.com/android/platform/superproject/+/master:packages/apps/PermissionController/src/com/android/packageinstaller/role/model/BrowserRoleBehavior.java
```
PEConn, line 105: Is this publicly accessible?

andreban: it is :D

PR #107, opened 2020-06-02
Add support for TWAs to launch on ARC++.

jeevan-shikaram:
Hi Peter and Andre,

This change allows for TWAs to launch on ARC++. I've made this change in the CCT fallback strategy in case ARC++ ends up implementing native TWA support.

Thanks!


PEConn:
The code looks fine to me - add a few lines for the description and we're good to go.

PEConn:
@andreban do you have anything to add? Otherwise I'm happy to merge.

jeevan-shikaram:
Using the package manager is the only way I know to reliably check if an APK is running on ARC++ unfortunately. I'll open up another pull request if I find a way.

I didn't feel that the launch was any slower, but I am running a fairly high sec device so I'm not sure how this will affect some lower spec devices.

PEConn:
Andre, we still want to check the PackageManager so we can be forward compatible in case ChromeOS CCT/TWA support improves and we can interact with the CustomTabsService the same way we would with Android.

PR #111, opened 2020-06-12
Port customtabs demo

NotWoods:
Adds an AndroidX version of the demo from custom-tabs-client


PEConn:
Hey Tiger, so this was copied from [the old GitHub repo right][1]? We ended up copying that over to live in the [Chromium source code here][2].

I'm unsure where it should live - or if it should live in both places. @andreban what do you think?

A few comments on the actual PR:

* I think it should be called "custom_tabs_example_app" or something a bit more descriptive than "customtabs".
* The readme needs to be changed.
* We should get rid of all of the browser actions related functionality - they've been deprecated (and Chrome (the only browser that ever supported them) doesn't support them anymore).
* Were there any non-mechanical changes you made? If so could you highlight them so I can just skim the rest.

Thanks,
Peter

[1]: https://github.com/GoogleChrome/custom-tabs-client/tree/master/demos
[2]: https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/android/examples/custom_tabs_client/src/

NotWoods:
Yup, this was copied from the old GitHub repo and classes were changes to their custom tabs equivalents. I switched the PNG drawables with vector drawables as well.

andreban:
android-browser-helper makes it easier to use Custom Tabs too, so I think we need demos. Probably not something for this PR, but we should replace code in the demos with android-browser-helper where possible (eg: `CustomTabsHelper` has a method used to detect compatible browsers, but the ABH implementation is more flexible)

andreban:
@NotWoods if you are building something with Custom Tabs and missing some API or have some feedback, I'd be happy to chat.

PEConn:
Skimming through the code it LGTM.

PR #113, opened 2020-06-18
Make the library compatible with Android 11

### LINKED ISSUE #112: Prepare for package visibility changes on Android 11
Applications targeting sdk level 30 will need to declare their queries in order to be able to read packages from the PackageManager. This impacts applications using Trusted Web Activity, as we need to query for installed browsers to connect to the service or decide on fallback strategies.

Adding a `queries` section with the appropriate markup android to the library `AndroidManifest.xml` will also import those queries into application using the library, via Manifest merging.

The following markup should be enough:

```xml
<queries>
    <intent>
        <action android:name="android.intent.action.VIEW" />
        <category android:name="android.intent.category.BROWSABLE" />
        <data android:scheme="https" />
    </intent>
</queries>
```
More information on package visibility changes on: https://developer.android.com/preview/privacy/package-visibility

andreban:
- Adds a `queries` section to the library `AndroidManifest.xml`
  that is automatically merged with the manifest for applications
  using the library.
- Updates Trusted Web Activity demos to target API level 30
- Doesn't update the Billing demo and the Custom Tabs demos. Those
  will be updated separately.

Closes #112

### DIFF
--- a/androidbrowserhelper/build.gradle
+++ b/androidbrowserhelper/build.gradle
@@ -18,11 +18,11 @@ apply plugin: 'com.android.library'
 apply plugin: 'maven-publish'
 
 android {
-    compileSdkVersion 28
+    compileSdkVersion 30
 
     defaultConfig {
         minSdkVersion 16
-        targetSdkVersion 28
+        targetSdkVersion 30
         versionCode 1
         versionName "1.3.1-alpha01"
 
--- a/androidbrowserhelper/src/main/AndroidManifest.xml
+++ b/androidbrowserhelper/src/main/AndroidManifest.xml
@@ -15,4 +15,12 @@
 -->
 
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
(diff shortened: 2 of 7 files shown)


androidbrowserhelper/build.gradle
```diff
@@ -18,11 +18,11 @@ apply plugin: 'com.android.library'
 apply plugin: 'maven-publish'
 
 android {
-    compileSdkVersion 28
+    compileSdkVersion 30
```
PEConn, line 21: How stable is this?

andreban: This will only affect devices that have Android 11. Otherwise, doesn't have any side-effects.

andreban:
> Did you have trouble trying to bump the compileSdkVersion of Play Billing?

Didn't actually try it yet - it *should* work, but I'm not sure I'll need any extra setup to test the billing app.

PR #114, opened 2020-06-18
Convert vector drawable to bitmap

NotWoods:
Ports `Drawable.toBitmap()` from the AndroidX kotlin extensions code, and uses it to serialize the vector back button.

Previously `BitmapFactory` just returned `null`. While a bitmap resource could be used again, I think this is better for sample code as most apps have moved to vectors.

Also deleted the reader mode icon since that was used by browser actions.


andreban:
@PEConn can you take a look?

PEConn:
LGTM

PR #117, opened 2020-06-25
Support location delegation in TWA

EiraGe:
1. Add the `ExtraCommandHandler` class for handling extraCommand in DelegationService
2. Add a `locationdelegation` module that includes `LocationDelegationExtraCommandHandler`, `PermissionRequestActivity`, `LocationProviderGmsCore` & `LocationProviderAndroid` for location delegations.


andreban:
@EiraGe I have just created an `alpha` branch: https://github.com/GoogleChrome/android-browser-helper/tree/alpha

EiraGe:
@andreban Oh Great!  Changed base.

androidbrowserhelper/build.gradle
```diff
@@ -43,13 +43,14 @@ android {
...
 dependencies {
-    api 'androidx.browser:browser:1.2.0'
+    api 'androidx.browser:browser:1.3.0-alpha03'
 
     implementation fileTree(dir: 'libs', include: ['*.jar'])
 
     implementation 'androidx.annotation:annotation:1.1.0'
     implementation 'androidx.appcompat:appcompat:1.0.2'
     implementation 'androidx.core:core:1.0.2'
+    implementation 'com.google.android.gms:play-services-location:17.0.0'
```
PEConn, line 53: Is there any way to make this optional? I'm wary of increasing the APK size for users who don't use this feature?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -14,19 +14,145 @@
...
     @Override
     public TokenStore getTokenStore() {
         return new SharedPreferencesTokenStore(this);
     }
+
+    @Nullable
+    @Override
+    public Bundle onExtraCommand(
+            String commandName, Bundle args, TrustedWebActivityCallbackRemote callback) {
+        Log.d(TAG, commandName);
```
PEConn, line 66: We probably want to remove this (or add more context).

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -14,19 +14,145 @@
...
+        Bundle result = new Bundle();
+        switch (commandName) {
+            case CHECK_LOCATION_PERMISSION_COMMAND_NAME: {
+                if (callback == null) break;
+
+                result.putBoolean(commandName, true);
```
PEConn, line 72: (I mentioned this to you in an email, here's probably a better place for it.)

How come we're using the commandName as a key for the success value? Couldn't we just use "success"?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -14,19 +14,145 @@
...
     }
+
+    @Nullable
+    @Override
+    public Bundle onExtraCommand(
+            String commandName, Bundle args, TrustedWebActivityCallbackRemote callback) {
+        Log.d(TAG, commandName);
+        Bundle result = new Bundle();
+        switch (commandName) {
+            case CHECK_LOCATION_PERMISSION_COMMAND_NAME: {
```
PEConn, line 69: I don't think you need curly braces for switch statements?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -14,19 +14,145 @@
...
+                break;
+            }
+            case STOP_LOCATION_COMMAND_NAME: {
+                stopLocationProvider();
+
+                result.putBoolean(commandName, true);
+                break;
+            }
+            default: {
+                result.putBoolean(commandName, false);
```
PEConn, line 91: So in some situations the results bundle contains false, and in some cases, it contains nothing at all (eg from the if statements on lines 70, 77 and 78.

Maybe could you just put `false` in the Bundle below line 67 then any happy paths will overwrite it?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -14,19 +14,145 @@
...
+            }
+        }
+        return result;
+    }
+
+    private boolean checkPermission(String permission) {
+        return ContextCompat.checkSelfPermission(this, permission) == PackageManager.PERMISSION_GRANTED;
+    }
+
+    private void requestPermission(TrustedWebActivityCallbackRemote callback, String permission) {
```
PEConn, line 101: Do you mind moving these methods into a separate class? We're going to be adding more stuff to the extraCommand logic (eg payments), so it would be nice if this class were pretty much just onExtraCommand and then the other logic was in different classes.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -14,19 +14,145 @@
...
+        Bundle result = new Bundle();
+        for (int i = 0; i < permissions.length; i++) {
+            if (TextUtils.equals(permissions[i], mLocationPermission)) {
+                result.putBoolean(LOCATION_PERMISSION_RESULT, grantedResult[i] == PackageManager.PERMISSION_GRANTED);
+            }
+        }
+        try {
+            callback.runExtraCallback(CHECK_LOCATION_PERMISSION_COMMAND_NAME, result);
+        } catch (RemoteException e) {
+            e.printStackTrace();
```
PEConn, line 135: Probably a `Log.e` instead?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -14,19 +14,145 @@
...
+                    data.getStringArray(EXTRA_PERMISSIONS), data.getIntArray(EXTRA_GRANT_RESULTS));
+            return true;
+        });
+        final Messenger messenger = new Messenger(handler);
+        handler.post(() -> {
+            Intent intent = new Intent(this, PermissionRequestActivity.class);
+            intent.putExtra(EXTRA_PERMISSIONS, new String[]{permission});
+            intent.putExtra(EXTRA_RESULT_RECEIVER, messenger);
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_NEW_DOCUMENT);
+            startActivity(intent);
```
PEConn, line 120: Usually what I do is have a static method on PermissionRequestActivity that launches it - so here you'd have:

```
PermissionRequestActivity.launch(context, permission, messenger);
```

And you can keep all the EXTRA keys private inside PermissionRequestActivity.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -14,19 +14,145 @@
...
+        }
+
+        Handler handler = new Handler(Looper.getMainLooper(), message -> {
+            Bundle data = message.getData();
+            notifyPermissionResult(callback,
+                    data.getStringArray(EXTRA_PERMISSIONS), data.getIntArray(EXTRA_GRANT_RESULTS));
+            return true;
+        });
+        final Messenger messenger = new Messenger(handler);
+        handler.post(() -> {
```
PEConn, line 115: Do you need to do this? Can you not call startActivity from a background thread?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -14,19 +14,145 @@
...
+            Intent intent = new Intent(this, PermissionRequestActivity.class);
+            intent.putExtra(EXTRA_PERMISSIONS, new String[]{permission});
+            intent.putExtra(EXTRA_RESULT_RECEIVER, messenger);
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_NEW_DOCUMENT);
+            startActivity(intent);
+        });
+    }
+
+    private void notifyPermissionResult(
```
PEConn, line 124: Can this method be static?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -14,19 +14,145 @@
...
+    private void requestPermission(TrustedWebActivityCallbackRemote callback, String permission) {
+        if (checkPermission(permission)) {
+            notifyPermissionResult(callback,
+                    new String[]{permission}, new int[]{PackageManager.PERMISSION_GRANTED});
+            return;
+        }
+
+        Handler handler = new Handler(Looper.getMainLooper(), message -> {
+            Bundle data = message.getData();
+            notifyPermissionResult(callback,
```
PEConn, line 110: It's kind of a shame we can't just pass the TrustedWebActivityCallbackRemote over to the Activity and let the Activity call it itself.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/PermissionRequestActivity.java
```diff
@@ -0,0 +1,56 @@
...
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        String[] permissions = getIntent().getStringArrayExtra(EXTRA_PERMISSIONS);
+        mMessenger = getIntent().getParcelableExtra(EXTRA_RESULT_RECEIVER);
+        ActivityCompat.requestPermissions(this, permissions, 0);
+    }
+
+    public void onRequestPermissionsResult(int requestCode,
```
PEConn, line 41: Should there be an `@Override` here?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/PermissionRequestActivity.java
```diff
@@ -0,0 +1,56 @@
...
+
+import android.os.Bundle;
+import android.os.Message;
+import android.os.Messenger;
+import android.os.RemoteException;
+
+import androidx.appcompat.app.AppCompatActivity;
+import androidx.core.app.ActivityCompat;
+
+public class PermissionRequestActivity extends AppCompatActivity {
```
PEConn, line 25: javadoc please

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LocationProviderGmsCore.java
```diff
@@ -0,0 +1,190 @@
...
+import androidx.browser.trusted.TrustedWebActivityCallbackRemote;
+
+/**
+ * This is a LocationProvider using Google Play Services.
+ * <p>
+ * https://developers.google.com/android/reference/com/google/android/gms/location/package-summary
+ */
+public class LocationProviderGmsCore
+        implements ConnectionCallbacks, OnConnectionFailedListener, LocationListener {
+    private static final String TAG = "TWA_LocationProviderGms";
```
PEConn, line 43: TAGs should be less than 20 characters.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -14,19 +14,145 @@
...
+        }
+        try {
+            callback.runExtraCallback(CHECK_LOCATION_PERMISSION_COMMAND_NAME, result);
+        } catch (RemoteException e) {
+            e.printStackTrace();
+        }
+    }
+
+    private void startLocationProvider(
+            final TrustedWebActivityCallbackRemote locationChangeCallback, final boolean enableHighAccuracy) {
```
PEConn, line 140: Does this need to be `final`?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LocationProviderGmsCore.java
```diff
@@ -0,0 +1,190 @@
...
+    private boolean mEnablehighAccuracy;
+    private LocationRequest mLocationRequest;
+
+    LocationProviderGmsCore(Context context) {
+        mGoogleApiClient = new GoogleApiClient.Builder(context)
+                .addApi(LocationServices.API)
+                .addConnectionCallbacks(this)
+                .addOnConnectionFailedListener(this)
+                .build();
+        assert mGoogleApiClient != null;
```
PEConn, line 67: How do we signal this case to Chrome? Does Chrome care?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LocationProviderGmsCore.java
```diff
@@ -0,0 +1,190 @@
...
+    private static LocationProviderGmsCore sProvider;
+    private final GoogleApiClient mGoogleApiClient;
+    private FusedLocationProviderApi mLocationProviderApi = LocationServices.FusedLocationApi;
+
+    private TrustedWebActivityCallbackRemote mCallback;
+
+    private boolean mEnablehighAccuracy;
+    private LocationRequest mLocationRequest;
+
+    LocationProviderGmsCore(Context context) {
```
PEConn, line 61: What happens if a web page requests location and doesn't "stop" it? And what happens if A page requests location multiple times? I'm wondering whether we need to be wary of any state left around.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LocationProviderGmsCore.java
```diff
@@ -0,0 +1,190 @@
...
+            mLocationProviderApi.requestLocationUpdates(
+                    mGoogleApiClient, mLocationRequest, this, Looper.getMainLooper());
+        } catch (IllegalStateException | SecurityException e) {
+            // IllegalStateException is thrown "If this method is executed in a thread that has not
+            // called Looper.prepare()". SecurityException is thrown if there is no permission, see
+            // https://crbug.com/731271.
+            Log.e(TAG, " mLocationProviderApi.requestLocationUpdates() " + e);
+            notifyLocationErrorWithMessage(
+                    "Failed to request location updates: " + e.toString());
+            assert false;
```
PEConn, line 112: Do we really want to crash here?

(Also, while asserts in Java are turned on in Chrome, I think in general they are not, so asserts in the Android Browser Helper will have no effect.)

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LocationProviderAndroid.java
```diff
@@ -0,0 +1,206 @@
...
+    private static LocationProviderAndroid sProvider;
+
+    public static LocationProviderAndroid getInstance() {
+        if (sProvider == null) {
+            sProvider = new LocationProviderAndroid();
+        }
+        return sProvider;
+    }
+
+    LocationProviderAndroid() {
```
PEConn, line 56: You don't need this - an empty, no parameter constructor is created automatically for you.

androidbrowserhelper/build.gradle
EiraGe: The only way I can think of is make it default to use the android location provider, and add another target in the demos/ that include the gms provider. 
The down side is in Chrome tab, default is using the gms one, and fallback to the android one if no play service available. So this may make TWA behaviour different with Chrome.
WDYT?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
EiraGe: I'm thinking that we can/should keep the basic method here: when these logic moving out of extraCommand, we still need the method in TWAService. I think it's nice to have the current structure align with what we meant to have in the future?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
EiraGe: :(

androidbrowserhelper/build.gradle
PEConn: So, I don't think we'd add something in demos, maybe top level (like `playbilling`) or maybe in some `features` directory or something. Then the TWA could depend on just the core `android-browser-helper` or add in the features it wants.

Ideally this core part would have no code to do with location delegation at all (or maybe just some stubs), and we would find some way to register them.

I'm thinking something like this:

1. The `DelegationService` has an interface like `ExtraCommandHandlers`.
1. The `DelegationService` has a list of these `ExtraCommandHandlers` and when it gets an extraCommand, it loops through them, seeing if any can deal with it.
1. You move all the code you added to `DelegationService` into the `LocationDelegationExtraCommandHandler`, which lives in a different module.
1. Somehow, the LocationDelegation module registers itself to the `DelegationService`.

That way we keep the code `android-browser-helper` module light and we can develop all these new feature modules in quite a neat, separate way. The only problem is that I haven't quite figured out how to do step 4 yet.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LocationProviderGmsCore.java
PEConn: Oh really? For some reason I thought in Chrome it was 20. Nevermind then.

androidbrowserhelper/build.gradle
EiraGe: I updated the PR to using `ExtraCommandHandler` and a separate `locationdelegation` module. Will add a demo and the update the bubblewrap change later

locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/PermissionRequestActivity.java
```diff
@@ -0,0 +1,114 @@
...
+import android.os.Message;
+import android.os.Messenger;
+import android.os.RemoteException;
+import android.text.TextUtils;
+
+import androidx.appcompat.app.AppCompatActivity;
+import androidx.browser.trusted.TrustedWebActivityCallbackRemote;
+import androidx.core.app.ActivityCompat;
+
+import com.google.androidbrowserhelper.trusted.DelegationService;
```
PEConn, line 33: Is this import needed?

locationdelegation/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+dependencies {
+    api 'androidx.browser:browser:1.2.0'
+
+    implementation fileTree(dir: "libs", include: ["*.jar"])
+    implementation 'androidx.appcompat:appcompat:1.1.0'
+    implementation 'com.google.android.gms:play-services-location:17.0.0'
+
+    testImplementation 'junit:junit:4.12'
+    androidTestImplementation 'androidx.test.ext:junit:1.1.1'
+    androidTestImplementation 'androidx.test.espresso:espresso-core:3.2.0'
```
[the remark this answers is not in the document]
EiraGe: Removed

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ExtraCommandHandler.java
```diff
@@ -0,0 +1,25 @@
+package com.google.androidbrowserhelper.trusted;
```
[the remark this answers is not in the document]
EiraGe: added copyright to all new files

locationdelegation/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
...
+#   public *;
+#}
+
+# Uncomment this to preserve the line number information for
+# debugging stack traces.
+#-keepattributes SourceFile,LineNumberTable
+
+# If you keep the line number information, uncomment this to
+# hide the original source file name.
+#-renamesourcefileattribute SourceFile
```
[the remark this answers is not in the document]
EiraGe: added new lines

locationdelegation/src/androidTest/java/com/google/androidbrowserhelper/locationdelegation/ExampleInstrumentedTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+import static org.junit.Assert.*;
+
+/**
+ * Instrumented test, which will execute on an Android device.
+ *
+ * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
+ */
+@RunWith(AndroidJUnit4.class)
+public class ExampleInstrumentedTest {
```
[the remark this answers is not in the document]
EiraGe: Removed all auto generated test files (will add real tests later)

locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/PermissionRequestActivity.java
EiraGe: Removed

locationdelegation/src/test/java/com/google/androidbrowserhelper/locationdelegation/ExampleUnitTest.java
```diff
@@ -0,0 +1,17 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+/**
+ * Example local unit test, which will execute on the development machine (host).
+ *
+ * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
+ */
+public class ExampleUnitTest {
```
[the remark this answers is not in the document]
EiraGe: Removed

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ExtraCommandHandler.java
```diff
@@ -0,0 +1,25 @@
...
+
```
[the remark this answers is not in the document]
EiraGe: Removed

locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/LocationDelegationExtraCommandHandler.java
```diff
@@ -0,0 +1,84 @@
...
+import androidx.browser.trusted.TrustedWebActivityCallbackRemote;
+
+import com.google.androidbrowserhelper.trusted.ExtraCommandHandler;
+
+public class LocationDelegationExtraCommandHandler implements ExtraCommandHandler {
+    static final String CHECK_LOCATION_PERMISSION_COMMAND_NAME = "checkAndroidLocationPermission";
+    private static final String START_LOCATION_COMMAND_NAME = "startLocation";
+    private static final String STOP_LOCATION_COMMAND_NAME = "stopLocation";
+
+    private boolean mIsProviderGmsCore;
```
PEConn, line 31: Instead of this, could you have a `private final LocationListener` and set it in the constructor (checking if the Play Services are available there)?

Or if the Play Services check could take some time, we could have a `getLocationListener()` method that initializes it lazily.

(Basically so that we don't have an `if (mIsProviderGmsCore)` in both `startLocationProvider` and `stopLocationProvider`.)

locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/PermissionRequestActivity.java
```diff
@@ -0,0 +1,112 @@
...
+    private static final String EXTRA_RESULT_RECEIVER = "EXTRA_RESULT_RECEIVER";
+    private static final String EXTRA_PERMISSIONS = "EXTRA_PERMISSIONS";
+    private static final String EXTRA_GRANT_RESULTS = "EXTRA_GRANTED_RESULT";
+
+    private static final String mLocationPermission = Manifest.permission.ACCESS_FINE_LOCATION;
+
+    private Messenger mMessenger;
+
+    /**
+     * Launches the {@link PermissionRequestActivity} to request location permission, and then send
```
PEConn, line 50: nit: send -> sends

locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/PermissionRequestActivity.java
```diff
@@ -0,0 +1,112 @@
...
+ * then finish.
+ */
+public class PermissionRequestActivity extends AppCompatActivity {
+    private static final String LOCATION_PERMISSION_RESULT = "locationPermissionResult";
+
+    private static final String EXTRA_RESULT_RECEIVER = "EXTRA_RESULT_RECEIVER";
+    private static final String EXTRA_PERMISSIONS = "EXTRA_PERMISSIONS";
+    private static final String EXTRA_GRANT_RESULTS = "EXTRA_GRANTED_RESULT";
+
+    private static final String mLocationPermission = Manifest.permission.ACCESS_FINE_LOCATION;
```
PEConn, line 45: nit: this should be `LOCATION_PERMISSION`

locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/PermissionRequestActivity.java
```diff
@@ -0,0 +1,112 @@
...
+import androidx.appcompat.app.AppCompatActivity;
+import androidx.browser.trusted.TrustedWebActivityCallbackRemote;
+import androidx.core.app.ActivityCompat;
+
+/**
+ * This is a simple transparent activity that will bring up the permission prompt. On either approve
+ * or disapprove, this will send the result via the {@link Messenger} provided with the intent, and
+ * then finish.
+ */
+public class PermissionRequestActivity extends AppCompatActivity {
```
PEConn, line 38: Do we not need to include this Activity in a manifest somewhere?

locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/LocationProviderAndroid.java
```diff
@@ -0,0 +1,203 @@
...
+        unregisterFromLocationUpdates();
+        mCallback = callback;
+        registerForLocationUpdates(context, enableHighAccuracy);
+    }
+
+    public void stop() {
+        unregisterFromLocationUpdates();
+    }
+
+    public boolean isRunning() {
```
PEConn, line 67: Does this need to be public?

locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/LocationProviderAndroid.java
```diff
@@ -0,0 +1,203 @@
...
+     * Registers this object with the location service.
+     */
+    private void registerForLocationUpdates(Context context, boolean enableHighAccuracy) {
+        createLocationManagerIfNeeded(context);
+        if (usePassiveOneShotLocation()) return;
+
+        assert !mIsRunning;
+        mIsRunning = true;
+
+        // We're running on the main thread. The C++ side is responsible to
```
PEConn, line 111: The C++ side of what? Of Chrome? Of the Android location SDK?

locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/LocationProviderGmsCore.java
```diff
@@ -0,0 +1,190 @@
...
+import com.google.android.gms.location.FusedLocationProviderApi;
+import com.google.android.gms.location.LocationListener;
+import com.google.android.gms.location.LocationRequest;
+import com.google.android.gms.location.LocationServices;
+
+import androidx.browser.trusted.TrustedWebActivityCallbackRemote;
+
+/**
+ * This is a LocationProvider using Google Play Services.
+ * <p>
```
PEConn, line 38: nit: We don't normally put `<p>`s in our javadoc

locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/LocationProviderGmsCore.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+        if (location.hasSpeed()) {
+            locationResult.putDouble("speed", location.getSpeed());
+        }
+
+        try {
+            mCallback.runExtraCallback(EXTRA_NEW_LOCATION_AVAILABLE_CALLBACK, locationResult);
+        } catch (RemoteException e) {
+            Log.e(TAG,"Caught RemoteException sending location update callback." );
+            stop();
```
PEConn, line 176: Can we share some of this code with the other LocationProvider?

locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/LocationProviderAndroid.java
EiraGe: done

locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/LocationProviderAndroid.java
EiraGe: Removed now. This comments applies to the locationProvider class in chromium repo, not our version.

locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/PermissionRequestActivity.java
PEConn: So I think we can add it to a manifest for the module, then if this module is included in some app, the manifest is merged together.

EiraGe: wow, I didn't realize I can add an empty <application> tag in the manifest. Added now

PR #121, opened 2020-06-29
Supports screenOrientation in TWAS

sclimwi:
1) Add a field in the LauncherActivity called android:screenOrientation to lock the splashScreen
2) Add a meta data in the AndroidManifest.xml in order to lock all the activities' into a lockType
3) Add a string field in LanucherActivityData called screenOrientation
4) Add the .setScreenOrientation() method for the twaActivityIntentBuilder, which will be implemented in the androidX part.

Still working in progress but I'd like to have reviews so I can change accordingly.

@glennhartmann @EiraGe @andreban @PEConn

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
@@ -154,6 +154,7 @@ public class LauncherActivity extends AppCompatActivity {
                         .setColorSchemeParams(
                                 CustomTabsIntent.COLOR_SCHEME_DARK, darkModeColorScheme)
                         .setDisplayMode(getDisplayMode());
+                       .setScreenOrientation(mMetadata.screenOrientation);
 
         if (mMetadata.additionalTrustedOrigins != null) {
             twaBuilder.setAdditionalTrustedOrigins(mMetadata.additionalTrustedOrigins);
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivityMetadata.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivityMetadata.java
@@ -20,6 +20,7 @@ import android.content.Context;
 import android.content.pm.PackageManager;
 import android.content.res.Resources;
 import android.os.Bundle;
+import android.util.Log;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -34,6 +35,8 @@ import static android.view.WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_M
  * Parses and holds on to metadata parameters associated with {@link LauncherActivity}.
  */
 public class LauncherActivityMetadata {
+
+    private static final String TAG = "LauncherActivityMetadata";
 
     /**
      * Url to launch in a Trusted Web Activity, unless other url provided in a VIEW intent.
@@ -114,6 +117,22 @@ public class LauncherActivityMetadata {
     private static final String METADATA_DISPLAY_MODE =
             "android.support.customtabs.trusted.DISPLAY_MODE";
 
+
+    /**
+     * The screen orientation to use when launching the Trusted Web Activity. Possible values are
+     * "any",
+     * "natural",
+     * "landscape",
+     * "landscape-primary",
+     * "landscape-secondary",
+     * "portrait",
+     * "portrait-primary",
+     * "portrait-secondary".
+     */
+    private static final String METADATA_SCREEN_ORIENTATION =
+            "android.support.customtabs.trusted.SCREEN_ORIENTATION";
+
+
     private final static int DEFAULT_COLOR_ID = android.R.color.white;
 
     @Nullable public final String defaultUrl;
@@ -128,6 +147,7 @@ public class LauncherActivityMetadata {
     @Nullable public final List<String> additionalTrustedOrigins;
     @Nullable public final String fallbackStrategyType;
     public final TrustedWebActivityDisplayMode displayMode;
+    public final String screenOrientation;
 
     private LauncherActivityMetadata(@NonNull Bundle metaData, @NonNull Resources resources) {
         defaultUrl = metaData.getString(METADATA_DEFAULT_URL);
@@ -152,6 +172,8 @@ public class LauncherActivityMetadata {
         }
         fallbackStrategyType = metaData.getString(METADATA_FALLBACK_STRATEGY);
         displayMode = getDisplayMode(metaData);
+        screenOrientation = metaData.getString(METADATA_SCREEN_ORIENTATION);
+        Log.d(TAG, " screenOrientation (" + screenOrientation + ").");
     }
 
     private static TrustedWebActivityDisplayMode getDisplayMode(@NonNull Bundle metaData) {
--- a/demos/twa-basic/src/main/AndroidManifest.xml
+++ b/demos/twa-basic/src/main/AndroidManifest.xml
@@ -26,7 +26,8 @@
             android:resource="@string/asset_statements" />
 
         <activity android:name="com.google.androidbrowserhelper.trusted.LauncherActivity"
-            android:label="@string/app_name">
+            android:label="@string/app_name"
+            android:screenOrientation="landscape">
             <meta-data android:name="android.support.customtabs.trusted.DEFAULT_URL"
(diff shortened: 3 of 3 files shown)


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -154,6 +154,7 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
                         .setColorSchemeParams(
                                 CustomTabsIntent.COLOR_SCHEME_DARK, darkModeColorScheme)
                         .setDisplayMode(getDisplayMode());
+                       .setScreenOrientation(mMetadata.screenOrientation);
```
PEConn, line 157: This method doesn't exist yet, right?

Also, nit: indentation

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivityMetadata.java
```diff
@@ -152,6 +172,8 @@ private LauncherActivityMetadata(@NonNull Bundle metaData, @NonNull Resources re
         }
         fallbackStrategyType = metaData.getString(METADATA_FALLBACK_STRATEGY);
         displayMode = getDisplayMode(metaData);
+        screenOrientation = metaData.getString(METADATA_SCREEN_ORIENTATION);
+        Log.d(TAG, " screenOrientation (" + screenOrientation + ").");
```
PEConn, line 176: nit: I'm guessing we're getting rid of this?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivityMetadata.java
```diff
@@ -114,6 +117,22 @@
...
+     * The screen orientation to use when launching the Trusted Web Activity. Possible values are
+     * "any",
+     * "natural",
+     * "landscape",
+     * "landscape-primary",
+     * "landscape-secondary",
+     * "portrait",
+     * "portrait-primary",
+     * "portrait-secondary".
+     */
```
PEConn, line 131: Are these values coming from the web world or from the Android world? (Either way, can you add a link to where they come from?)

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivityMetadata.java
```diff
@@ -114,6 +117,22 @@
...
+    private static final String METADATA_SCREEN_ORIENTATION =
+            "android.support.customtabs.trusted.SCREEN_ORIENTATION";
+
+
```
PEConn, line 135: nit: one newline too many

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivityMetadata.java
```diff
@@ -114,6 +117,22 @@
     private static final String METADATA_DISPLAY_MODE =
             "android.support.customtabs.trusted.DISPLAY_MODE";
 
+
```
PEConn, line 120: nit: one newline too many

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivityMetadata.java
```diff
@@ -128,6 +147,7 @@
     @Nullable public final List<String> additionalTrustedOrigins;
     @Nullable public final String fallbackStrategyType;
     public final TrustedWebActivityDisplayMode displayMode;
+    public final String screenOrientation;
```
PEConn, line 150: This should also be `@Nullable`, right?

demos/twa-basic/src/main/AndroidManifest.xml
```diff
@@ -49,6 +50,10 @@
             <meta-data android:name="android.support.customtabs.trusted.DISPLAY_MODE"
                 android:value="immersive"/>
 
+
+            <meta-data android:name="android.support.customtabs.trusted.SCREEN_ORIENTATION"
+                android:value="landscape"/>
```
PEConn, line 55: I'm wondering if we actually need this. Maybe we could just read the screen orientation of `LauncherActivity` and then forward that. I can't see any reason for the LauncherActivity and the Chrome Activity to be different orientations.

andreban: Can we add the screen orientation bit to a separate demo? Most developers won't care about a specific orientation and I'd like to avoid folks copying / pasting twa-basic and getting a fixed orientation by accident.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
andreban: `mMetadata.screenOrientation` will be null here, if the property is not defined in `AndroidManifest.xml`.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivityMetadata.java
andreban: +1 to `@Nullable` - another option would be to ensure setting a default value (`"any"`) when calling `screenOrientation = metaData.getString(METADATA_SCREEN_ORIENTATION);`. I'd prefer `@Nullable` though, as this would mean allowing the "default" behaviour provided by `androix.browser` or the browser implementation itself. We should only call `.setScreenOrientation(mMetadata.screenOrientation);` when this is not null (see previous comment).

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
sclimwi: @PEConn The setScreenOrientation is not existed yet, we need to add that for the twaIntentBuilder class in androidX.

demos/twa-basic/src/main/AndroidManifest.xml
sclimwi: @PEConn. Here is a template we should add for the LauncherActivity. Indeed we need to read the orientation info from web manifest, then pass that into this launcherActivity and also to the meta data. The reason we have to have this is that without this android:screenOrientation setting in the AndroidManifest.xml, the launcherActivity will be launched with a portrait mode first, then rotating to landscape if we set that for later activities. This was the bug we fixed for the WebAPKs.

sclimwi: @andreban Thanks for your suggestion. We can make a separate demo for the orientation implementation without affecting others.

PEConn: So at the moment we have two pieces of information in the AndroidManifest:

1. We have the metadata that tells us to launch the browser in landscape.
2. We have `android:screenOrientation="landscape"` on the Activity.

What I'm suggesting is that we get rid of the first one, then when launching the browser, instead of checking for the LauncherActivityMetadata, we just check to see whether we are currently in landscape or not.

sclimwi: My understanding is that we may want to keep both. 

1. https://stuff.mit.edu/afs/sipb/project/android/docs/guide/topics/manifest/activity-element.html#screen. The android:screenOrientation has a different set of values from Web screenOrientation: https://www.w3.org/TR/screen-orientation/#screenorientation-interface

2. We can detect the screenOrientation inside the activity, but we cannot differentiate from the the natural portrait mode or the locked portrait mode when the app is launching with a portrait. Some apps want to lock the screen as portrait as well, and this is the case for Pinterest:   crbug.com/812797. We can get the screenOrientation value this way: https://www.tutorialspoint.com/how-can-i-get-the-current-screen-orientation-in-android.

PEConn: Ah, they're both good points, thanks for the explanation.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivityMetadata.java
sclimwi: This is from the w3.org spec: https://www.w3.org/TR/screen-orientation/#screenorientation-interface.

Will add this link to the new patch.

andreban:
Closing this in favour of #122

PR #122, opened 2020-07-01
Supports screenOrientation in TWAS

sclimwi:
1) Add a field in the LauncherActivity called android:screenOrientation to lock the splashScreen
2) Add a meta data in the AndroidManifest.xml in order to lock all the activities' into a lockType
3) Add a @Nullable string field in LanucherActivityData called screenOrientation
4) Add the .setScreenOrientation() method for the twaActivityIntentBuilder, which will be implemented in the androidX part
5) Add a new demo called twa-nofification-delegation-orientation so we can test the orientation locking without changing other demos

I think I did not do the properly, but this version looks cleaner. 
**I need to figure out how to push to the same PR, instead of creating new PR every time (:.** 

 @PEConn @andreban @glennhartmann @EiraGe

PR #127, opened 2020-07-15
screenOrientation support for TWAs

sclimwi:
The previous PR (#122 ) is having too many unnecessary files so I made a new branch with the implementation of screenOrientation support.

This one should be clean and right to get reviews. PTAL @PEConn @andreban @glennhartmann @EiraGe

Thanks very much in advance!


sclimwi:
Hey, All. 

I modified the twa-demo to test on a simple website with res from twa-basic, instead of relying the old twa-notification-delegation files.

It has a conflict with the alpha but I think screen orientation locking has to rely on this `androidx.browser:browser:1.3.0-alpha05`.

PTAL and let me know if there are issues to be taken care of, @PEConn @andreban @glennhartmann @EiraGe.

build.gradle
```diff
@@ -34,6 +33,7 @@ buildscript {
 
 allprojects {
     repositories {
+        maven { url '/home/scli/more-android-sdk/extra/m2repository' }
```
PEConn, line 36: What's this about?

demos/twa-basic/src/main/AndroidManifest.xml
```diff
@@ -49,6 +50,10 @@
             <meta-data android:name="android.support.customtabs.trusted.DISPLAY_MODE"
                 android:value="immersive"/>
 
+
```
PEConn, line 53: nit: Random newline.

demos/twa-basic/src/main/AndroidManifest.xml
```diff
@@ -49,6 +50,10 @@
...
+            <meta-data android:name="android.support.customtabs.trusted.SCREEN_ORIENTATION"
+                android:value="landscape"/>
```
PEConn, line 55: So I do like the idea of having an example for how to set the value in our basic demo, but I don't think that I want the `twa-basic` demo to launch in landscape. Could you just change this to portrait please?

demos/twa-orientation/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,102 @@
...
+                android:resource="@drawable/splash"/>
+
+            <meta-data android:name="android.support.customtabs.trusted.SPLASH_SCREEN_BACKGROUND_COLOR"
+                android:resource="@color/backgroundColor"/>
+
+            <meta-data android:name="android.support.customtabs.trusted.SPLASH_SCREEN_FADE_OUT_DURATION"
+                android:value="300"/>
+
+            <meta-data android:name="android.support.customtabs.trusted.FILE_PROVIDER_AUTHORITY"
+                android:value="@string/provider_authority"/>
```
PEConn, line 60: So in this block, pretty much only the default url and the screen orientation meta-data are necessary. If you leave the rest out, you'll lose the splash screen and end up with default colours.

I don't mind leaving the rest in (to make the demo look pretty) or taking them out to make it obvious which are the useful bits of the code. Up to you.

build.gradle
sclimwi: That is the dependency on my local build. Will remove it shortly.

demos/twa-basic/src/main/AndroidManifest.xml
sclimwi: Ack.

demos/twa-basic/src/main/AndroidManifest.xml
sclimwi: Sure.

demos/twa-orientation/src/main/AndroidManifest.xml
sclimwi: I think for the testing purpose, moving them out would be good. I will remove these lines.

sclimwi:
Hey, @PEConn and @andreban. I resolved the conflicts. Could you please merge this PR.

humiston:
Is there any time estimation for when this functionality will be available via release?

andreban:
@humiston this change depends on an alpha version of `androidx.browser` to become stable. We expect that to happen in a couple of months.

MrNeilRobbins:
Looking forward to this feature. Glad to see it's being worked on. Our TWA needs to be locked to landscape.

I've cloned the repo and tested the demo in the 'alpha' branch BUT the lock doesn't seem to work? When I rotate my device the app rotates. The demo app points to HTML that has a button to enable full screen and lock orientation with JavaScript and that works. But the TWA orientation lock doesn't work. Is there something I'm missing when running the demo.

Glad to help debug.

sclimwi:
Hey Neil,

In order to lock the orientation, the app has to be in 1) TWA mode (the
search bar is not shown), and 2) we have to rely on the chromium build that
has the orientation support code added. I think these are the two most
important pieces in order to lock the orientation.

The demo app there is to show the locking through JS will take precedence
over the default orientation we added.

Best
SL

EiraGe:
@MrNeilRobbins Thanks for testing it. As @sclimwi mentioned, you need to download Chrome Canary and set it as default browser to use the feature. (The code is in Chrome M86 and should roll out to Chrome stable around Mid Oct)

MrNeilRobbins:
Thanks. 

I've installed Chrome Canary on my device (Chrome Canary 86.0.4232.0) OS Android 10; SM-T720 Build/QP1A.190711.020

Then went to:
`Settings > Apps > ... > Default Apps > Browser app > set to Chrome Canary`

But when run **twa-orientation** and it launches on the device the app still rotates freely without locking the orientation.

sclimwi:
@MrNeilRobbins Thanks for checking out that. I do not know if you run through this step yet but just in case:

> Enable debug mode
> Open Chrome on the development device, navigate to chrome://flags, search for an item called Enable command line on non-rooted devices and change it to ENABLED and then restart the browser.
> Next, on the Terminal application of your operating system, use the Android Debug Bridge (installed with Android Studio), and run the following command:
> adb shell "echo '_ --disable-digital-asset-link-verification-for-url=\"https://airhorner.com\"' > /data/local/tmp/chrome-command-line"
> 
> Close Chrome and re-launch your application from Android Studio. The application should now be shown in full-screen.

from this document: https://developers.google.com/web/android/trusted-web-activity/integration-guide. 

This step is to make sure the website is in TWA mode. You may want to change the command to be 

`adb shell "echo '_ --disable-digital-asset-link-verification-for-url=\"https://tulip-forest-limpet.glitch.me\"' > /data/local/tmp/chrome-command-line" `

One more note is that sometime even we set the default browser, the app will still be launched with default Chrome version. We want to make sure we could pick up the right Chrome version. 

Hopefully with the above steps, you can get the orientation up and running.

shunxing:
I tried to set the value to portrait in my manifest but it ultimately doesn't work. 
image
When rotating my simulator, the app still rotates. Can someone explain me whether I did something wrong ? @sclimwi @PEConn

PR #129, opened 2020-07-22
Create trusted-web-activity-browser-support.md

andreban:
Kicks off a document mapping Trusted Web Activity support across browsers.

### DIFF
--- a/docs/trusted-web-activity-browser-support.md
+++ b/docs/trusted-web-activity-browser-support.md
@@ -1,10 +1,9 @@
 Trusted Web Activity Browser Support
 ====================================
 
-**Important:** This is an unnofficial list of Trusted Web Activity support across
-browsers. The list is not exhaustive and is maintained by the community.
-Please file an [issue][14] if the information is incorrect, incomplete of if you'd
-like to see other browsers added.
+**Important:** We do our best to keep this up to date - if you spot anything missing or
+incorrect, please file an [issue][14].
+
 
 |Browser          |Basic Support              |Splash Screen|Notification Delegation|
 |:---------------:|:-------------------------:|:-----------:|:---------------------:|
--- a/docs/trusted-web-activity-browser-support.md
+++ b/docs/trusted-web-activity-browser-support.md
@@ -1,0 +1,42 @@
+Trusted Web Activity Browser Support
+====================================
+
+**Important:** This is an unnofficial list of Trusted Web Activity support across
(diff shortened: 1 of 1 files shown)


docs/trusted-web-activity-browser-support.md
```diff
@@ -0,0 +1,42 @@
+Trusted Web Activity Browser Support
+====================================
+
+**Important:** This is an unnofficial list of Trusted Web Activity support across
```
PEConn, line 4: If we're in a repo owned by "GoogleChrome", can this really be unofficial?

Why are we saying it's unofficial? In case there are mistakes?

docs/trusted-web-activity-browser-support.md
```diff
@@ -0,0 +1,42 @@
...
+browsers. The list is not exhaustive and is maintained by the community.
+Please file an [issue][14] if the information is incorrect, incomplete of if you'd
```
PEConn, line 6: File an issue or a pull request? Are we (most likely you) going to end up manually verifying everything?

docs/trusted-web-activity-browser-support.md
andreban: We can definitely keep the Chrome details updated, but not other vendors. Maybe we can word this differently?

docs/trusted-web-activity-browser-support.md
andreban: An issue sounds like a good first step? PRs are definitely welcome, but issues may be easier to file and have a quick chat. (I definitely don't plan to manually verify everything)

docs/trusted-web-activity-browser-support.md
PEConn: How about "We do our best to keep this up to date - if you spot anything missing or incorrect, please file an issue at XX".

docs/trusted-web-activity-browser-support.md
PEConn: Would it be worth having an Issue template for this specifically?

docs/trusted-web-activity-browser-support.md
andreban: Updated

docs/trusted-web-activity-browser-support.md
andreban: Might be - I'll waiting until we start getting issues on this and then we can map what would be useful to have on the template.

PR #131, opened 2020-07-23
Checks for CrOS TWA requirements

andreban:
- Checks the application for the requirements of running on Chrome OS.
- Prints log messages when those are missing.

The existence of `web-manifest-json` and the `web-manifest-url` meta tag indicates the app is intended to be used as a Trusted Web Activity in Chrome OS. If one of the two is not available, this PR makes the app print a message to the log and doesn't add the `EXTRA_LAUNCH_AS_TRUSTED_WEB_ACTIVITY` extra.

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
@@ -54,7 +54,8 @@ public class TwaLauncher {
             intent.intent.setPackage(providerPackage);
         }
         // Add the TWA flag to the intent if the app is running on ARC++ on Chrome OS.
-        if (context.getPackageManager().hasSystemFeature(ARC_FEATURE)) {
+        if (context.getPackageManager().hasSystemFeature(ARC_FEATURE) &&
+                Utils.hasChromeOsSupport(context)) {
             intent.intent.putExtra(TrustedWebUtils.EXTRA_LAUNCH_AS_TRUSTED_WEB_ACTIVITY, true);
         }
         intent.launchUrl(context, twaBuilder.getUri());
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
@@ -52,7 +52,7 @@ public class TwaLauncher {
             intent.intent.setPackage(providerPackage);
         }
         // Add the TWA flag to the intent if the app is running on ARC++ on Chrome OS.
-        if (Utils.hasChromeOsSupport(context)) {
+        if (Utils.shouldAddExtraForChromeOs(context)) {
             intent.intent.putExtra(TrustedWebUtils.EXTRA_LAUNCH_AS_TRUSTED_WEB_ACTIVITY, true);
         }
         intent.launchUrl(context, twaBuilder.getUri());
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
@@ -43,8 +43,6 @@ public class TwaLauncher {
 
     private static final int DEFAULT_SESSION_ID = 96375;
 
-    private static final String ARC_FEATURE = "org.chromium.arc";
-
     public static final FallbackStrategy CCT_FALLBACK_STRATEGY =
             (context, twaBuilder, providerPackage, completionCallback) -> {
         // CustomTabsIntent will fall back to launching the Browser if there are no Custom Tabs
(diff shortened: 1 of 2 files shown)


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/Utils.java
```diff
@@ -99,4 +102,41 @@ public static Bitmap convertDrawableToBitmap(Context context, int drawableId) {
...
-}
\ No newline at end of file
+
+    /**
+     * Verifies if the application has the required meta-tag, `web_manifest_url`, and the
+     * `web_app_manifest.json` file, required to enable the Trusted Web Activity on Chrome OS.
+     * Important: The content the meta-tag and the JSON fire are not validated.
+     */
+    public static boolean hasChromeOsSupport(Context context) {
+        // We first check if /app/res/web_app_manifest.json is available
```
PEConn, line 112: nit: You're missing full stops on comments (in a few other places as well).

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/Utils.java
```diff
@@ -99,4 +102,41 @@ public static Bitmap convertDrawableToBitmap(Context context, int drawableId) {
...
+        int id = context.getResources().getIdentifier(
+                "web_app_manifest.json", "raw", context.getPackageName());
+        // If an attribute with that name doesn't exist the result is 0 (which is an invalid ID for
+        // attributes)
+        if (id == 0) {
+            Log.e("ChromeOS-Support", "Could not find '/res/raw/web_app_manifest.json'");
```
PEConn, line 118: I think this qualifies as a warning rather than an error - aren't there valid cases where TWAs don't want to do this on ChromeOS?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/Utils.java
```diff
@@ -99,4 +102,41 @@ public static Bitmap convertDrawableToBitmap(Context context, int drawableId) {
...
+            if (webManifestUrl == null) {
+                Log.e("ChromeOS-Support",
+                        "Could not the 'web_manifest_url' meta-tag in 'AndroidManifest.xml'.'");
+            }
+            context.getPackageManager().getResourcesForApplication(context.getPackageName());
+            return false;
+        } catch (PackageManager.NameNotFoundException ex) {
+            // This should never happen, as we're retrieving info for the app itself, but we log
+            // and allow returning true nevertheless.
+            Log.e("ChromeOS-Support", "Error retrieving application meta-data");
```
PEConn, line 136: nit: full stops on logs

PEConn, line 136: You want a `return false` here.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/Utils.java
```diff
@@ -99,4 +102,41 @@ public static Bitmap convertDrawableToBitmap(Context context, int drawableId) {
...
+        try {
+            ApplicationInfo applicationInfo = context.getPackageManager()
+                    .getApplicationInfo(context.getPackageName(), PackageManager.GET_META_DATA);
+            String webManifestUrl = applicationInfo.metaData.getString("web_manifest_url");
+            if (webManifestUrl == null) {
+                Log.e("ChromeOS-Support",
+                        "Could not the 'web_manifest_url' meta-tag in 'AndroidManifest.xml'.'");
+            }
+            context.getPackageManager().getResourcesForApplication(context.getPackageName());
+            return false;
```
PEConn, line 132: Do you want to return false here - this is the happy case? The logic seems a bit off.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/Utils.java
andreban: Added.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -54,7 +54,8 @@
             intent.intent.setPackage(providerPackage);
         }
         // Add the TWA flag to the intent if the app is running on ARC++ on Chrome OS.
-        if (context.getPackageManager().hasSystemFeature(ARC_FEATURE)) {
+        if (context.getPackageManager().hasSystemFeature(ARC_FEATURE) &&
+                Utils.hasChromeOsSupport(context)) {
```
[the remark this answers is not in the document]
andreban: done.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/Utils.java
andreban: Agreed - changed to warnings.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/Utils.java
andreban: done.

andreban: done.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/Utils.java
andreban: Yeah, changed this a bit. 

I was planning to add code to check for the actual content of the meta-tag while making the draft (then, obviously forgot about it when making this an actual PR).

I'm just checking for `bundle.containsKey()` now instead retrieving the value and returning false if the key is not there. Otherwise, return true.

andreban:
Closing this, as it is stale.

PR #134, opened 2020-07-27
Create TWA payments demo.

PEConn:
Creates a demo that responds to the Digital Goods API added recently to Chrome.

It currently produces canned responses (it doesn't talk to Play Billing) and doesn't contain a Payment Provider - both of these are future work.


PEConn:
Quite a bit of this code will be moved into the `playbilling` module, probably overwriting the `bundles` package.

PEConn:
Also, the Java classes are the interesting ones - the rest are all pretty standard (although I did replace the auto-generated icon with a pretty, open-licensed one I found).

demos/twa-play-billing/src/main/java/com/google/androidbrowserhelper/demos/playbilling/BillingDelegationService.java
```diff
@@ -0,0 +1,41 @@
...
+    public static final String KEY_SUCCESS = "success";
+
+    @Nullable
+    @Override
+    public Bundle onExtraCommand(@NonNull String commandName, @NonNull Bundle args,
+                                 @Nullable TrustedWebActivityCallbackRemote callbackRemote) {
+        Bundle bundle = new Bundle();
+        bundle.putBoolean(KEY_SUCCESS, false);
+
+        switch (commandName) {
```
andreban, line 21: If I understood correctly so far, this bit of the code will receive the messages from the browser (get detail and acknowledge) and dispatch to the appropriate call implementation. The call implementation will handle the conversation with the billing API. 

I feel `GetDatailsCall` and `AcknowledgeCall` are similar and should implement a common interface. May be wrong though.

demos/twa-play-billing/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,70 @@
...
+    <application
+        android:allowBackup="true"
+        android:icon="@mipmap/ic_launcher"
+        android:label="@string/app_name"
+        android:roundIcon="@mipmap/ic_launcher_round"
+        android:supportsRtl="true"
+        android:theme="@style/AppTheme"
+        tools:ignore="GoogleAppIndexingWarning">
+
+        <activity
```
andreban, line 27: MainActivity is empty and does nothing for now, but I'm guessing this is on purpose at the moment.

settings.gradle
```diff
@@ -1,5 +1,11 @@
...
+include ':playbilling'
+
 include ':demos:custom-tabs-example-app'
 include ':demos:payment-provider'
+include ':demos:twa-basic'
+include ':demos:twa-custom-launcher'
+include ':demos:twa-multi-domain'
 include ':demos:twa-notification-delegation'
+include ':demos:twa-play-billing'
+include ':demos:twa-webview-fallback'
```
EiraGe, line 11: nit: newline at eof?

demos/twa-play-billing/src/main/res/values/ic_launcher_background.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <color name="ic_launcher_background">#FFFFFF</color>
+</resources>
```
EiraGe, line 4: nit: add newline?

demos/twa-play-billing/src/main/java/com/google/androidbrowserhelper/demos/playbilling/AcknowledgeCall.java
```diff
@@ -0,0 +1,55 @@
...
+            return null;
+        }
+
+        String token = args.getString(PARAM_ACKNOWLEDGE_PURCHASE_TOKEN);
+        boolean makeAvailableAgain = args.getBoolean(PARAM_ACKNOWLEDGE_MAKE_AVAILABLE_AGAIN);
+
+        return new AcknowledgeCall(token, makeAvailableAgain, callback);
+    }
+
+    public void respond(int responseCode) {
```
EiraGe, line 45: Can this use the BillingClient.BillingResponseCode intdef?

demos/twa-play-billing/src/main/java/com/google/androidbrowserhelper/demos/playbilling/AcknowledgeCall.java
```diff
@@ -0,0 +1,55 @@
...
+    private static final String PARAM_ACKNOWLEDGE_MAKE_AVAILABLE_AGAIN =
+            "acknowledge.makeAvailableAgain";
+    private static final String RESPONSE_ACKNOWLEDGE = "acknowledge.response";
+    private static final String RESPONSE_ACKNOWLEDGE_RESPONSE_CODE = "acknowledge.responseCode";
+
+    public final String purchaseToken;
+    public final boolean makeAvailableAgain;
+    private final TrustedWebActivityCallbackRemote mCallback;
+
+    private AcknowledgeCall(String purchaseToken, boolean makeAvailableAgain,
```
padr31, line 22: I am missing some documentation on why this needs to be private and created through create(...) instead of a public constructor. It seems like the class is not a singleton.

demos/twa-play-billing/src/main/java/com/google/androidbrowserhelper/demos/playbilling/AcknowledgeCall.java
```diff
@@ -0,0 +1,55 @@
...
+        return new AcknowledgeCall(token, makeAvailableAgain, callback);
+    }
+
+    public void respond(int responseCode) {
+        try {
+            Bundle args = new Bundle();
+            args.putInt(RESPONSE_ACKNOWLEDGE_RESPONSE_CODE, responseCode);
+            mCallback.runExtraCallback(RESPONSE_ACKNOWLEDGE, args);
+        } catch (RemoteException e) {
+            // TODO: Something...
```
padr31, line 51: NIT: Maybe left this in unintentionally?

demos/twa-play-billing/src/main/java/com/google/androidbrowserhelper/demos/playbilling/BillingDelegationService.java
```diff
@@ -0,0 +1,41 @@
+package com.google.androidbrowserhelper.demos.playbilling;
```
padr31, line 1: Missing license header, few other files as well

demos/twa-play-billing/src/main/java/com/google/androidbrowserhelper/demos/playbilling/GetDetailsCall.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public void respond(int responseCode, ItemDetails... itemDetails) {
+        try {
+            Bundle args = new Bundle();
+            args.putInt(RESPONSE_GET_DETAILS_RESPONSE_CODE, responseCode);
+            args.putParcelableArray(RESPONSE_GET_DETAILS_DETAILS_LIST,
+                    toParcelableArray(itemDetails));
+            mCallback.runExtraCallback(RESPONSE_GET_DETAILS, args);
+        } catch (Exception e) {
+            // TODO: Something...
```
padr31, line 44: NIT: Same as previously

demos/twa-play-billing/src/main/java/com/google/androidbrowserhelper/demos/playbilling/AcknowledgeCall.java
PEConn: So I don't really have too strong of a reason why this constructor is private, other than it doesn't need to be public at the moment. But I do have a reason why I made the static `create` method (instead of having that logic in a constructor).

We're creating the object from the `args` bundle, which may or may not be valid. If we had a constructor that was given an invalid `args` (for example, if it doesn't contain a purchase token) we could either:
- Construct an `AcknowledgeCall` that is incomplete (eg, `purchaseToken` is null).
- Throw an exception.

On the other hand, we use a static `create` method, we have another option - we can return `null`.

It's better if `AcknowledgeCall` can't represent invalid values - either "it exists and its valid" or "it doesn't exist". The difference between the other two is mostly ergonomics - would we prefer the caller to have to use a try-catch or an `if (thing == null)`, but also a bit of performance (Java exceptions have some overhead).

So I would definitely prefer to have a `create(Bundle args)` over a `new AcknowledgeCall(Bundle args)`. However, if there are other things that construct this object, I would be fine also having a public `new Acknowledge(String purchaseToken, boolean makeAvailableAgain)`.

PEConn:
This pull request was split into multiple smaller ones and each of them have been landed:
#150 #154 #155 #159 #161 #166 #167 #168 #175

PR #136, opened 2020-07-29
Create demo for location delegation

EiraGe:
A demo for making use of :locationdelegation module

@andreban @PEConn PTAL Thanks!


demos/twa-location-delegation/README.md
```diff
@@ -0,0 +1,4 @@
+# Trusted Web Activity / Location Delegation Demo
+
+This demo application shows how to make the TWA app uses Android location permission instead of
+browser permission.
```
andreban, line 4: Can we add some pointers to the chunks that are relevant to location delegation? It looks to me the necessary steps for a developer who wants to enable this are:

1. Adding the location delegation dependency on build.gradle
2. Creating the LocationDelegationService
3. Adding a reference to LocationDelegationService on the AndroidManifest.xml

EiraGe: Sounds good. Added.

demos/twa-location-delegation/README.md
```diff
@@ -0,0 +1,7 @@
+# Trusted Web Activity / Location Delegation Demo
+
+This demo application shows how to make the TWA app uses Android location permission instead of
```
PEConn, line 3: nits:
uses -> use
Android ->Android's
browser -> the browser's

demos/twa-location-delegation/README.md
```diff
@@ -0,0 +1,7 @@
...
+browser permission:
+1. Adding the location delegation dependency on build.gradle
```
PEConn, line 5: Could you put full stops at the end of these points please?

demos/twa-location-delegation/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,99 @@
...
+            android:name="androidx.core.content.FileProvider"
+            android:authorities="@string/provider_authority"
+            android:grantUriPermissions="true"
+            android:exported="false">
+            <meta-data
+                android:name="android.support.FILE_PROVIDER_PATHS"
+                android:resource="@xml/filepaths" />
+        </provider>
+
+        <service
```
PEConn, line 88: Maybe move this up to the top? Or have some comments saying "This is the interesting part" and "This is the boring part that's the same for all TWAs".

demos/twa-location-delegation/src/main/java/com/google/browser/examples/twa_location_delegation/LocationDelegationService.java
```diff
@@ -0,0 +1,25 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.browser.examples.twa_location_delegation;
+
+import com.google.androidbrowserhelper.trusted.DelegationService;
+import com.google.androidbrowserhelper.locationdelegation.LocationDelegationExtraCommandHandler;
+
+public class LocationDelegationService extends DelegationService {
```
PEConn, line 20: I'm not sure this should be called "LocationDelegationService" - yes, what this does is location delegation, but the way we've designed the API is that multiple different features can go in here. Eg, if someone wanted to add the Payments Service in as well, they'd just add a line in the constructor.

Maybe call this ExtraFeaturesService or OptionalFeaturesService or something? (Open to better name suggestions.)

demos/twa-location-delegation/README.md
EiraGe: Done

demos/twa-location-delegation/src/main/java/com/google/browser/examples/twa_location_delegation/LocationDelegationService.java
EiraGe: Renamed to ExtraFeaturesService

demos/twa-location-delegation/README.md
EiraGe: Done.

demos/twa-location-delegation/src/main/AndroidManifest.xml
EiraGe: Moved to top and added some comments.

EiraGe:
> tip: when updating the PR, using a regular commit/push instead of force-push allows the reviewer to check the individual commit that addresses the review feedback, which is a lot easier than re-reviewing the whole PR.

Ah, I see. I was thinking that squashing into one can make the merged commit history more clean. Will do regular commit next time. Thanks  :)

andreban:
> Ah, I see. I was thinking that squashing into one can make the merged commit history more clean. Will do regular commit next time. Thanks :)

We don't usually do it, but Github has an option to squash when merging - we can get the best of both worlds.

PR #141, opened 2020-07-30
Add TWA Quality Enforcement

EiraGe:
Use extraCallbackWithResult in CustomTabsCallback to handle quality enforcement message from the browser.
When gets a "Notify" message, logs the error (for the initial version) and return.
When gets a "Crash" message, post a RuntimeException and return.

@PEConn PTAL Thanks


EiraGe:
Updated to using the delegate interface as we discussed last week. @PEConn PTAL Thanks :)

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/QualityEnforcer.java
```diff
@@ -0,0 +1,76 @@
...
+import android.util.Log;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.browser.customtabs.CustomTabsCallback;
+
+/**
+ * This class handles the quality enforcement messages from the browser to enforce the quality bar
+ * on the websites shown inside Trusted Web Activities.
+ * Browser should send a message when violation occurs. for example if a link from a page on the
```
PEConn, line 29: nit: Capitalize the "for".

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/QualityEnforcer.java
```diff
@@ -0,0 +1,76 @@
...
+    static final String KEY_CRASH_REASON = "crash_reason";
+    static final String KEY_SUCCESS = "success";
+
+    private final Delegate mDelegate;
+
+    interface Delegate {
+        void crash(String message);
+    }
+
+    public QualityEnforcer() {
```
PEConn, line 49: Can you add javadoc to this, the other constructor, `Delegate` and the `crash` method on it?

For this constructor, something like "Constructor to be used in prod that throws a RuntimeException" and for the other constructor "Constructor for use in tests" are good.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/QualityEnforcer.java
```diff
@@ -0,0 +1,76 @@
...
+        mDelegate = (message) -> {
+            new Handler(Looper.getMainLooper()).post(() -> {
```
PEConn, line 51: Could you add a comment about why we're putting this on the Looper?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/QualityEnforcer.java
```diff
@@ -0,0 +1,76 @@
...
+    QualityEnforcer(Delegate delegate) { mDelegate = delegate; }
+
+    @Nullable
+    @Override
+    public Bundle extraCallbackWithResult(
+            @NonNull String callbackName, @Nullable Bundle args) {
+        Bundle result = new Bundle();
+        result.putBoolean(KEY_SUCCESS, false);
+        if (callbackName.equals(NOTIFY)) {
+            String message = (args != null) ? args.getString(KEY_CRASH_REASON) : "";
```
PEConn, line 66: You're doing this line in both branches - you can move it out of the `if`.

Also, I think if the crash reason is missing or empty, we should return success = false and do nothing.

androidbrowserhelper/src/androidTest/java/com/google/androidbrowserhelper/trusted/TwaQualityEnforcerTest.java
```diff
@@ -0,0 +1,121 @@
...
+            TestBrowser.class,
+            TestCustomTabsServiceSupportsTwas.class,
+            TestCustomTabsService.class
+    );
+    @Rule
+    public final ActivityTestRule<TestActivity> mActivityTestRule =
+            new ActivityTestRule<>(TestActivity.class, false, true);
+
+    @Rule
+    public final ExpectedException thrown = ExpectedException.none();
```
PEConn, line 72: Is this needed? (If so, `mThrown` please.)

EiraGe:
Remove "quality_enforcement.notify" to sync with Chromium change.

@PEConn  PTAL Thanks

EiraGe:
@andreban could you take a look (and help merge it if it looks ok to you Thanks  :)

andreban (approved):
LGTM, Thank you! Sorry for the delay.

PR #148, opened 2020-08-05
Add site settings shortcut to supporting twa's

padr31:
The commit adds a dynamic shortcut that links to the TWA's site settings provided that the following conditions are met:
    1. The manifest of the TWA declared ManageDataLauncherActivity - needed to launch site settings
    2. The TWA provider supports site settings either via a custom manifest category, or if it is a legacy version of chrome

### DIFF
--- a/androidbrowserhelper/src/androidTest/AndroidManifest.xml
+++ b/androidbrowserhelper/src/androidTest/AndroidManifest.xml
@@ -16,7 +16,31 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           package="android.support.customtabs">
     <uses-sdk android:minSdkVersion="16"/>
-    <application>
+    <queries>
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+
+            <category android:name="android.intent.category.BROWSABLE" />
+
+            <data android:scheme="https" />
+        </intent>
+        <intent>
+            <action android:name="android.support.customtabs.action.CustomTabsService" />
+        </intent>
+    </queries>
+
+    <application
+        android:manageSpaceActivity="com.google.androidbrowserhelper.trusted.ManageDataLauncherActivity">
+
+        <activity android:name="com.google.androidbrowserhelper.trusted.ManageDataLauncherActivity">
+            <meta-data
+                android:name="android.support.customtabs.trusted.MANAGE_SPACE_URL"
+                android:value="https://www.test.com/default_url/" />
+            <intent-filter>
+                <action android:name="android.support.customtabs.action.ACTION_MANAGE_TRUSTED_WEB_ACTIVITY_DATA"/>
+            </intent-filter>
+        </activity>
+
         <activity android:name="com.google.androidbrowserhelper.trusted.testcomponents.TestActivity"
                   android:enabled="false">
             <!-- A browsable intent filter is required for the DelegationService. -->
@@ -47,6 +71,7 @@
                 <action android:name="android.support.customtabs.action.CustomTabsService" />
                 <category android:name="androidx.browser.trusted.category.TrustedWebActivities" />
                 <category android:name="androidx.browser.trusted.category.TrustedWebActivitySplashScreensV1" />
+                <category android:name="androidx.browser.trusted.category.LaunchSiteSettings" />
             </intent-filter>
         </service>
 
--- a/androidbrowserhelper/src/androidTest/AndroidManifest.xml
+++ b/androidbrowserhelper/src/androidTest/AndroidManifest.xml
@@ -36,9 +36,6 @@
             <meta-data
                 android:name="android.support.customtabs.trusted.MANAGE_SPACE_URL"
                 android:value="https://www.test.com/default_url/" />
-            <intent-filter>
-                <action android:name="android.support.customtabs.action.ACTION_MANAGE_TRUSTED_WEB_ACTIVITY_DATA"/>
-            </intent-filter>
         </activity>
 
         <activity android:name="com.google.androidbrowserhelper.trusted.testcomponents.TestActivity"
@@ -108,6 +105,12 @@
             <meta-data
                 android:name="android.support.customtabs.trusted.STATUS_BAR_COLOR"
                 android:resource="@color/status_bar_color" />
+
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN"/>
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
         </activity>
 
         <!-- For TWA splash screens -->
--- a/androidbrowserhelper/src/androidTest/AndroidManifest.xml
+++ b/androidbrowserhelper/src/androidTest/AndroidManifest.xml
@@ -16,18 +16,6 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           package="android.support.customtabs">
     <uses-sdk android:minSdkVersion="16"/>
-    <queries>
-        <intent>
-            <action android:name="android.intent.action.VIEW" />
-
-            <category android:name="android.intent.category.BROWSABLE" />
-
-            <data android:scheme="https" />
-        </intent>
-        <intent>
-            <action android:name="android.support.customtabs.action.CustomTabsService" />
-        </intent>
-    </queries>
 
     <application
         android:manageSpaceActivity="com.google.androidbrowserhelper.trusted.ManageDataLauncherActivity">
--- a/androidbrowserhelper/src/androidTest/AndroidManifest.xml
+++ b/androidbrowserhelper/src/androidTest/AndroidManifest.xml
@@ -17,8 +17,7 @@
           package="android.support.customtabs">
     <uses-sdk android:minSdkVersion="16"/>
 
-    <application
-        android:manageSpaceActivity="com.google.androidbrowserhelper.trusted.ManageDataLauncherActivity">
+    <application>
 
         <activity android:name="com.google.androidbrowserhelper.trusted.ManageDataLauncherActivity">
             <meta-data
--- a/androidbrowserhelper/src/androidTest/java/com/google/androidbrowserhelper/trusted/LauncherActivityTest.java
+++ b/androidbrowserhelper/src/androidTest/java/com/google/androidbrowserhelper/trusted/LauncherActivityTest.java
@@ -118,7 +118,6 @@ public class LauncherActivityTest {
         assertEquals(SITE_SETTINGS_SHORTCUT_ID, shortcutManager.getDynamicShortcuts().get(0).getId());
     }
 
-
     private void checkColor(TestBrowser browser) {
         int requestedColor = browser.getIntent()
                 .getIntExtra(CustomTabsIntent.EXTRA_TOOLBAR_COLOR, 0);
--- a/androidbrowserhelper/src/androidTest/java/com/google/androidbrowserhelper/trusted/LauncherActivityTest.java
+++ b/androidbrowserhelper/src/androidTest/java/com/google/androidbrowserhelper/trusted/LauncherActivityTest.java
@@ -14,19 +14,16 @@
 
 package com.google.androidbrowserhelper.trusted;
 
-import static com.google.androidbrowserhelper.trusted.testutils.TestUtil.getBrowserActivityWhenLaunched;
+import static com.google.androidbrowserhelper.trusted.ManageDataLauncherActivity.SITE_SETTINGS_SHORTCUT_ID;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 
 import static androidx.browser.customtabs.TrustedWebUtils.EXTRA_LAUNCH_AS_TRUSTED_WEB_ACTIVITY;
-
 import android.content.Context;
-import android.content.Intent;
 import android.content.pm.ShortcutManager;
 import android.net.Uri;
 
 import com.google.androidbrowserhelper.test.R;
-import com.google.androidbrowserhelper.trusted.testcomponents.TestActivity;
 import com.google.androidbrowserhelper.trusted.testutils.EnableComponentsTestRule;
 import com.google.androidbrowserhelper.trusted.testcomponents.TestBrowser;
 import com.google.androidbrowserhelper.trusted.testcomponents.TestCustomTabsService;
@@ -38,10 +35,8 @@ import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
-import org.mockito.Mockito;
 
 import androidx.browser.customtabs.CustomTabsIntent;
-import androidx.browser.trusted.TrustedWebActivityIntentBuilder;
 import androidx.test.InstrumentationRegistry;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 import androidx.test.filters.MediumTest;
@@ -119,14 +114,10 @@ public class LauncherActivityTest {
     public void addsSiteSettingsShortcut() {
         TestBrowser browser = launch();
         ShortcutManager shortcutManager = mContext.getSystemService(ShortcutManager.class);
-        shortcutManager.getDynamicShortcuts();
-        Uri url = Uri.parse("https://padr31.github.io/");
-
-        TrustedWebActivityIntentBuilder builder = new TrustedWebActivityIntentBuilder(url);
-        Runnable launchRunnable = () -> new TwaLauncher(mActivityTestRule.getActivity()).launch(builder, null, null);
-        Intent intent = getBrowserActivityWhenLaunched(launchRunnable).getIntent();
-        //browser.get
+        assertEquals(1, shortcutManager.getDynamicShortcuts().size());
+        assertEquals(SITE_SETTINGS_SHORTCUT_ID, shortcutManager.getDynamicShortcuts().get(0).getId());
     }
+
 
     private void checkColor(TestBrowser browser) {
         int requestedColor = browser.getIntent()
--- a/androidbrowserhelper/src/androidTest/java/com/google/androidbrowserhelper/trusted/LauncherActivityTest.java
+++ b/androidbrowserhelper/src/androidTest/java/com/google/androidbrowserhelper/trusted/LauncherActivityTest.java
@@ -14,15 +14,19 @@
 
 package com.google.androidbrowserhelper.trusted;
 
+import static com.google.androidbrowserhelper.trusted.testutils.TestUtil.getBrowserActivityWhenLaunched;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 
 import static androidx.browser.customtabs.TrustedWebUtils.EXTRA_LAUNCH_AS_TRUSTED_WEB_ACTIVITY;
 
 import android.content.Context;
+import android.content.Intent;
+import android.content.pm.ShortcutManager;
 import android.net.Uri;
 
 import com.google.androidbrowserhelper.test.R;
+import com.google.androidbrowserhelper.trusted.testcomponents.TestActivity;
 import com.google.androidbrowserhelper.trusted.testutils.EnableComponentsTestRule;
 import com.google.androidbrowserhelper.trusted.testcomponents.TestBrowser;
 import com.google.androidbrowserhelper.trusted.testcomponents.TestCustomTabsService;
@@ -34,8 +38,10 @@ import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.mockito.Mockito;
 
 import androidx.browser.customtabs.CustomTabsIntent;
+import androidx.browser.trusted.TrustedWebActivityIntentBuilder;
 import androidx.test.InstrumentationRegistry;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 import androidx.test.filters.MediumTest;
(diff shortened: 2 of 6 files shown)


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -113,6 +118,8 @@
     @Nullable
     private TwaLauncher mTwaLauncher;
 
+    private PackageManager mPackageManager;
```
PEConn, line 121: This doesn't need to be a member. (You can still make it a local variable in `onCreate if you like.)

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -172,6 +179,30 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
 
         new TwaSharedPreferencesManager(this)
                 .writeLastLaunchedProviderPackageName(mTwaLauncher.getProviderPackage());
+
+        mPackageManager = getPackageManager();
```
PEConn, line 183: All of the new code here should go into its own method. Maybe even a method on some other class if we want to make it more easily testable.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -172,6 +179,30 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+        // Check whether ManageDataLauncherActivity is present in manifest.
+        Intent siteSettingsIntent = new Intent(this, ManageDataLauncherActivity.class);
+        siteSettingsIntent.setAction(ManageDataLauncherActivity.ACTION_MANAGE_TRUSTED_WEB_ACTIVITY_DATA);
```
PEConn, line 186: Do we need the action if we're creating the Intent with the class name?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -172,6 +179,30 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+        List<ResolveInfo> activities = mPackageManager.queryIntentActivities(siteSettingsIntent,
+                PackageManager.MATCH_DEFAULT_ONLY);
+        boolean siteSettingsAvailable = activities.size() > 0;
+
+        // Check whether TWA provider package supports site settings.
+        String packageName = mTwaLauncher.getProviderPackage();
+        boolean supportsTWASiteSettings = ManageDataLauncherActivity.packageSupportsSiteSettings(packageName, mPackageManager);
+
+        if(siteSettingsAvailable && supportsTWASiteSettings) {
```
PEConn, line 195: If we move this into a different function, we can simplify the control flow/unnest some of the code a bit.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -172,6 +179,30 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
```
PEConn, line 196: The version check should probably go right at the top of the function - if we can't set shortcuts, there's no reason to have checked all the stuff before.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -172,6 +179,30 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+                ShortcutManager shortcutManager = getSystemService(ShortcutManager.class);
+                ShortcutInfo shortcut = new ShortcutInfo.Builder(this, "siteSettings")
+                        .setShortLabel("Site Settings")
+                        .setLongLabel("Manage website notifications, permissions, etc.")
```
PEConn, line 200: We want to get the strings out of Resources if possible, this way the developer can override them (and they can be localized) - look at ChromeUpdatePrompt.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -172,6 +179,30 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+                        .setIntent(siteSettingsIntent)
+                        .build();
+                shortcutManager.setDynamicShortcuts(Collections.singletonList(shortcut));
+            }
+        }
```
PEConn, line 205: It's a much smaller case, but if we previously had a shortcut and now our browser no-longer supports them (eg, if the user changed their default browser), we should remove the old shortcut.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivity.java
```diff
@@ -294,4 +295,21 @@ public void onCustomTabsServiceConnected(ComponentName componentName,
         @Override
         public void onServiceDisconnected(ComponentName componentName) {}
     }
+
+    public static boolean packageSupportsSiteSettings(String packageName, PackageManager packageManager) {
+        if (packageName != null) {
```
PEConn, line 300: Turn this check into a guard clause (`if (packageName == null) return false;`).

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivity.java
```diff
@@ -294,4 +295,21 @@ public void onCustomTabsServiceConnected(ComponentName componentName,
...
+            if (ChromeLegacyUtils.supportsTrustedWebActivities(packageManager, packageName)) {
+                // Chrome 72-74 support site settings but don't yet have the category
```
PEConn, line 302: This comment implies that Chrome 75 has the category *for Site Settings*. (I'm guessing you copied this from the other code where it was referring to the category for TWAs in general.)

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -172,6 +179,30 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+        // Check whether ManageDataLauncherActivity is present in manifest.
+        Intent siteSettingsIntent = new Intent(this, ManageDataLauncherActivity.class);
+        siteSettingsIntent.setAction(ManageDataLauncherActivity.ACTION_MANAGE_TRUSTED_WEB_ACTIVITY_DATA);
+        List<ResolveInfo> activities = mPackageManager.queryIntentActivities(siteSettingsIntent,
+                PackageManager.MATCH_DEFAULT_ONLY);
+        boolean siteSettingsAvailable = activities.size() > 0;
+
+        // Check whether TWA provider package supports site settings.
+        String packageName = mTwaLauncher.getProviderPackage();
+        boolean supportsTWASiteSettings = ManageDataLauncherActivity.packageSupportsSiteSettings(packageName, mPackageManager);
```
PEConn, line 193: nit: line wrapping

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivity.java
```diff
@@ -294,4 +295,21 @@ public void onCustomTabsServiceConnected(ComponentName componentName,
...
+                return true;
+            } else {
+                Intent customTabsIntent = new Intent(CustomTabsService.ACTION_CUSTOM_TABS_CONNECTION);
+                customTabsIntent.addCategory(CATEGORY_LAUNCH_SITE_SETTINGS);
+                customTabsIntent.setPackage(packageName);
+                List<ResolveInfo> services = packageManager.queryIntentServices(customTabsIntent, PackageManager.GET_RESOLVED_FILTER);
```
PEConn, line 308: nit: line wrapping

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
padr31: Yes, otherwise we receive a nullpointer with the message: intent's action must be set

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -172,6 +176,40 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+        addSiteSettingsShortcut();
+    }
+
+    /**
+     * Adds dynamic shortcut to site settings if the twa provider and android version supports it.
+     *
+     * Removes previously added site settings shortcut if it is no longer supported, e.g. the user
+     * changed their default browser.
+     */
+    private void addSiteSettingsShortcut() {
```
PEConn, line 189: Do you think this method would be better in ManageDataLauncherActivity?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivity.java
```diff
@@ -77,6 +79,12 @@
     public static final String ACTION_MANAGE_TRUSTED_WEB_ACTIVITY_DATA =
             "android.support.customtabs.action.ACTION_MANAGE_TRUSTED_WEB_ACTIVITY_DATA";
 
+    public static final String SITE_SETTINGS_SHORTCUT_ID =
+            "android.support.customtabs.action.ACTION_MANAGE_TRUSTED_WEB_ACTIVITY_DATA";
```
PEConn, line 83: We should probably change the value of this (ending with something like "SITE_SETTINGS_SHORTCUT"?)

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivity.java
```diff
@@ -294,4 +302,55 @@ public void onCustomTabsServiceConnected(ComponentName componentName,
...
+     *
+     * Otherwise returns null if {@link ManageDataLauncherActivity} is not launchable
+     * or if shortcuts are not supported by the Android SDK version.
+     */
+    public static ShortcutInfo getSiteSettingsShortcutOrNull(Context context, PackageManager packageManager) {
+        Intent siteSettingsIntent = new Intent(context, ManageDataLauncherActivity.class);
+        siteSettingsIntent.setAction(ACTION_MANAGE_TRUSTED_WEB_ACTIVITY_DATA);
+        List<ResolveInfo> activities = packageManager.queryIntentActivities(siteSettingsIntent,
+                PackageManager.MATCH_DEFAULT_ONLY);
+        if(activities.size() > 0) {
```
PEConn, line 341: You can change this into a guard clause (`if (activities.size() == 0) return null`).

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivity.java
```diff
@@ -294,4 +302,55 @@ public void onCustomTabsServiceConnected(ComponentName componentName,
...
+
+    /**
+     * Returns the {@link ShortcutInfo} for a dynamic shortcut into site settings,
+     * provided that {@link ManageDataLauncherActivity} is present in the manifest
+     * and an Intent for managing site settings is available.
+     *
+     * Otherwise returns null if {@link ManageDataLauncherActivity} is not launchable
+     * or if shortcuts are not supported by the Android SDK version.
+     */
+    public static ShortcutInfo getSiteSettingsShortcutOrNull(Context context, PackageManager packageManager) {
```
PEConn, line 336: Can you mark the method as `@Nullable`?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivity.java
```diff
@@ -294,4 +302,55 @@ public void onCustomTabsServiceConnected(ComponentName componentName,
...
+        Intent siteSettingsIntent = new Intent(context, ManageDataLauncherActivity.class);
+        siteSettingsIntent.setAction(ACTION_MANAGE_TRUSTED_WEB_ACTIVITY_DATA);
+        List<ResolveInfo> activities = packageManager.queryIntentActivities(siteSettingsIntent,
+                PackageManager.MATCH_DEFAULT_ONLY);
+        if(activities.size() > 0) {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
+                return new ShortcutInfo.Builder(context, SITE_SETTINGS_SHORTCUT_ID)
+                        .setShortLabel("Site Settings")
```
PEConn, line 344: We should use string resources here (although we'll need to check whether or not they're included, look at how ChromeUpdatePrompt does it).

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivity.java
```diff
@@ -294,4 +302,55 @@ public void onCustomTabsServiceConnected(ComponentName componentName,
...
+                        .setLongLabel("Manage website notifications, permissions, etc.")
+                        .setIcon(Icon.createWithResource(context,
+                                android.R.drawable.ic_menu_preferences))
```
PEConn, line 347: This is good as a placeholder, but we should have some UX review to figure out the proper icon to use.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivity.java
```diff
@@ -294,4 +302,55 @@ public void onCustomTabsServiceConnected(ComponentName componentName,
...
+     * it is a legacy version of chrome that supports site settings but does not define
+     * the site settings category yet.
+     */
+    public static boolean packageSupportsSiteSettings(String packageName,
+            PackageManager packageManager) {
+        if (packageName == null) return false;
+        if (ChromeLegacyUtils.supportsSiteSettings(packageManager, packageName)) {
+            // Legacy Chrome supports site settings but does not have the site settings category.
+            return true;
+        } else {
```
PEConn, line 318: You don't need the `else` here since we're `return`ing above.

androidbrowserhelper/src/test/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivityTest.java
```diff
@@ -0,0 +1,203 @@
...
+import org.robolectric.RuntimeEnvironment;
+import org.robolectric.annotation.Config;
+import org.robolectric.annotation.internal.DoNotInstrument;
+import org.robolectric.shadows.ShadowPackageManager;
+
+import androidx.browser.customtabs.CustomTabsService;
+import androidx.test.InstrumentationRegistry;
+
+/**
+ * Tests for {@link TwaProviderPicker}.
```
PEConn, line 54: The class name is `ManageDataLauncherActivityTest`, but the javadoc says the tests are for `TwaProviderPicker`?

androidbrowserhelper/src/test/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivityTest.java
```diff
@@ -0,0 +1,203 @@
...
+                .addCategory(Intent.CATEGORY_BROWSABLE);
+
+        ResolveInfo resolveInfo = new ResolveInfo();
+        resolveInfo.activityInfo = new ActivityInfo();
+        resolveInfo.activityInfo.packageName = packageName;
+
+        mShadowPackageManager.addResolveInfoForIntent(intent, resolveInfo);
+    }
+
+    private void installCustomTabsProvider(String packageName) {
```
PEConn, line 143: Maybe it's worth sharing code between this and `TwaProviderPickerTest` - it looks like you've just copied it without any changes.

androidbrowserhelper/src/test/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivityTest.java
```diff
@@ -0,0 +1,203 @@
...
+        installChrome(CHROME_71_VERSION);
+        TwaProviderPicker.Action action = TwaProviderPicker.pickProvider(mPackageManager);
+
+        assertFalse(ManageDataLauncherActivity.packageSupportsSiteSettings(action.provider, mPackageManager));
+    }
+
+    @Test
+    public void browserWithCategorySupportsSiteSettings() {
+        installTrustedWebActivityProviderWithSiteSettingsCategory(TWA_PROVIDER_PACKAGE);
+
```
PEConn, line 93: nit: very small nit, but in the above two tests you don't have any newline between the "install..." and the "pickProvider" lines, whereas in these two tests you do.

androidbrowserhelper/src/test/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivityTest.java
```diff
@@ -0,0 +1,203 @@
...
+        TwaProviderPicker.Action action = TwaProviderPicker.pickProvider(mPackageManager);
+
+        assertTrue(ManageDataLauncherActivity.packageSupportsSiteSettings(action.provider, mPackageManager));
+    }
+
+    @Test
+    public void browserWithoutCategorySupportsSiteSettings() {
```
PEConn, line 100: This test should probably be called `browserWithoutCategoryDoesNotSupportSiteSettings`.

androidbrowserhelper/src/androidTest/AndroidManifest.xml
```diff
@@ -16,7 +16,28 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           package="android.support.customtabs">
     <uses-sdk android:minSdkVersion="16"/>
-    <application>
+    <queries>
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+
```
PEConn, line 22: nit: random newlines.

androidbrowserhelper/src/androidTest/AndroidManifest.xml
```diff
@@ -16,7 +16,28 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           package="android.support.customtabs">
     <uses-sdk android:minSdkVersion="16"/>
-    <application>
+    <queries>
```
PEConn, line 19: Now that we've got everything working - are we sure we need these?

androidbrowserhelper/src/androidTest/AndroidManifest.xml
```diff
@@ -83,6 +105,12 @@
             <meta-data
                 android:name="android.support.customtabs.trusted.STATUS_BAR_COLOR"
                 android:resource="@color/status_bar_color" />
+
+            <intent-filter>
```
PEConn, line 109: We should probably put a comment somewhere that this is required for the shortcuts to work.

androidbrowserhelper/src/androidTest/java/com/google/androidbrowserhelper/trusted/LauncherActivityTest.java
```diff
@@ -109,6 +110,15 @@ public void customTabHasStatusBarColor() {
...
 
+    @Test
+    public void addsSiteSettingsShortcut() {
+        TestBrowser browser = launch();
+        ShortcutManager shortcutManager = mContext.getSystemService(ShortcutManager.class);
+        assertEquals(1, shortcutManager.getDynamicShortcuts().size());
+        assertEquals(SITE_SETTINGS_SHORTCUT_ID, shortcutManager.getDynamicShortcuts().get(0).getId());
+    }
+
+
```
PEConn, line 121: nit: random newline

androidbrowserhelper/src/androidTest/AndroidManifest.xml
padr31: Good point, they are not needed anymore

androidbrowserhelper/src/androidTest/AndroidManifest.xml
```diff
@@ -16,7 +16,16 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           package="android.support.customtabs">
     <uses-sdk android:minSdkVersion="16"/>
-    <application>
+
+    <application
+        android:manageSpaceActivity="com.google.androidbrowserhelper.trusted.ManageDataLauncherActivity">
```
PEConn, line 21: I guess we don't really need this for the tests.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivity.java
```diff
@@ -332,25 +334,65 @@ public static boolean packageSupportsSiteSettings(String packageName,
...
-                        .setLongLabel("Manage website notifications, permissions, etc.")
-                        .setIcon(Icon.createWithResource(context,
-                                android.R.drawable.ic_menu_preferences))
-                        .setIntent(siteSettingsIntent)
-                        .build();
-            }
+
+        if(activities.size() == 0) return null;
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
```
PEConn, line 352: We can invert this check and then move it to the top of the method - there's no point querying the package manager if we know we can't create a shortcut.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivity.java
```diff
@@ -332,25 +334,65 @@ public static boolean packageSupportsSiteSettings(String packageName,
...
      * Otherwise returns null if {@link ManageDataLauncherActivity} is not launchable
      * or if shortcuts are not supported by the Android SDK version.
+     *
+     * The shortcut returned does not specify an activity. Thus when the shortcut is added,
+     * the app's main activity will be used by default. This activity needs to define the
+     * MAIN action and LAUNCHER category in order to attach the shortcut.
      */
-    public static ShortcutInfo getSiteSettingsShortcutOrNull(Context context, PackageManager packageManager) {
+    @Nullable
+    public static ShortcutInfo getSiteSettingsShortcutOrNull(Context context,
```
PEConn, line 343: Can this be private now? (Or do we want to keep it package-private so we can test it?)

padr31: I changed it to package private now

PR #152, opened 2020-08-14
Remove usages of androidx.appcompat

EiraGe:
Removes `androidx.appcompat` imports from android-browser-helper to
reduce the default TWA apk size (~500KB difference).

All activities in `androidbrowserhelper` module now extends `Activity`
instead of `AppCompatActivity`.

Most of the demo app's theme changed to `Theme.Translucent.NoTitleBar`
instead of the `Theme.LauncherActivity` that inherits the appcompat
style `Theme.AppCompat.NoActionBar`.
`Theme.Translucent.NoTitleBar` has the same translucent style as the
`Theme.LauncherActivity`, there should be no change in styles.

Some of the themes changed from `Theme.AppCompat.Light.DarkActionBar`
to `Theme.Material.Light.DarkActionBar`

The only demo app that's still import `androidx.appcompat` is the
`custom-tabs-example-app`. It contains many Activities and some of them
are using methods from `AppCompatActivity`, do not change it for now.

@PEConn @andreban  PTAL Thanks :)


andreban:
Wow!

I'm playing with this and the Bubblewrap PR locally and generated a project for http://pwa-directory.appspot.com/:

- Original size: **1.1Mb**
- With this PR: **333K**

Amazing!

PEConn:
As far as I know, AppCompatActivity was about ensuring Activities looked appropriately Material Designed on older phones. Since we don't care too much about the appearance of our demos, and since the Activities we use in prod are generally meant to be invisible, I don't see and issue with this.

PR #154, opened 2020-08-17
Add Play Billing Data classes

PEConn:
This change adds some fairly straightforward classes that will be used for communicating with the browser for the Digital Goods API.


playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/AcknowledgeCall.java
```diff
@@ -0,0 +1,54 @@
...
+
+import android.os.Bundle;
+
+import androidx.annotation.Nullable;
+
+/**
+ * A class for parsing Digital Goods API calls from the browser and converting them into a format
+ * suitable for calling the Play Biling library.
+ */
+public class AcknowledgeCall {
```
andreban, line 11: This may be overkill, so feel free to ignore,  but the similarity between `AcknowledgeCall` and `GetDetailsCall` make me feel like they have a common shared interface. 

eg:
```java
interface BillingCall<T> {
    void respond(int responseCode, T details) {
}
```

then 

```java
class AcknowledgeCall implements BillingCall<Void>
```
and 

```java
class GetDetailsCall implements BilllingCall<ItemDetails[]>
```

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/AcknowledgeCall.java
```diff
@@ -0,0 +1,54 @@
+package com.google.androidbrowserhelper.playbilling.digitalgoods;
```
andreban, line 1: We're missing license headers on this and other files.

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/AcknowledgeCall.java
PEConn: Yeah, I was considering that, but ultimately I couldn't see where we would use that interface. If an opportunity becomes apparent we can add it.

PR #157, opened 2020-08-17
Add divider color functionality

padr31:
Add support for changing navbar divider color in activities launched using LauncherActivity

Users required the ability to change the color of the navigation bar
divider custom tabs, which in some cases looked inconsistent. After adding
this feature to chromium and androidX, this PR adds it into android-browser-helper.
The activities launched with LauncherActivity will automatically set the divider color
by looking at manifest metadata. These are usually set by bubblewrap.

Test: Tested the integration of bubblewrap and android-browser-helper 
by building an example app and changing the navbar color.


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivityMetadata.java
```diff
@@ -126,12 +140,15 @@
             "android.support.customtabs.trusted.SCREEN_ORIENTATION";
 
     private final static int DEFAULT_COLOR_ID = android.R.color.white;
+    private final static int DEFAULT_DIVIDER_COLOR_ID = 0x00000000;
```
PEConn, line 143: Isn't `0x00000000` a colour, not a colour id? If you try to load from id 0, won't we get a crash?

There apparently is `android.R.color.transparent`.

andreban: As an additional note, `0` is not a valid resource Id. See https://developer.android.com/reference/android/content/res/Resources.html#getIdentifier(java.lang.String,%20java.lang.String,%20java.lang.String)

PEConn (approved):

PR #162, opened 2020-08-21
Chrome OS Play billing support

jeevan-shikaram:
Hello,

This change will allow the Chrome OS system package `org.chromium.arc.payment_app` to query a payment activity without being the package that launched it by skipping the verification process.

Thanks,
Jeevan


playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/PaymentVerifier.java
```diff
@@ -31,6 +31,9 @@
     // TODO: Should this be an instance class (eg PaymentStrategy)?
     // It would allow developers to override verification behaviour more easily.
 
+    private static final String ARC_FEATURE = "org.chromium.arc";
```
andreban, line 34: This is also defined on the main library: https://github.com/GoogleChrome/android-browser-helper/blob/master/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java#L46

Since billing depends on that library, I'm wondering if we should move the variable to an appropriate and make it accessible by the billing library so we can re-use.

jeevan-shikaram: I've create the support class. Please take a look again!

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ChromeOsSupport.java
```diff
@@ -0,0 +1,35 @@
...
+    /** The ARC++ package that quries for payments on behalf of TWAs on Chrome OS */
+    public static final String ARC_PAYMENT_APP = "org.chromium.arc.payment_app";
+
+    /**
+     * Checks if the application is running on ARC++.
+     *
+     * @param pm A {@link PackageManager}.
+     * @return True if running on ARC++.
+     */
+    public static boolean isRunningOnArc (PackageManager pm) {
```
andreban, line 32: nit: extra space between method name and parenthesis.
```suggestion
    public static boolean isRunningOnArc(PackageManager pm) {
```

andreban (approved):
LGTM, one small nit.

andreban (approved):
LGTM, Thank you!

jeevan-shikaram:
@PEConn Friendly ping

PEConn (approved):
Sorry for the delay, LGTM

PR #166, opened 2020-09-07
Implement DigitalGoods Acknowledge call

PEConn:
The Acknowledge call allows the TWA's website to trigger either the Play Billing acknowledge or consumeAsync methods. This is the last big part of DigitalGoodsRequestHandler to be implemented (although that class has some further cleanup still needed).

### DIFF
--- a/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/AcknowledgeCall.java
+++ b/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/AcknowledgeCall.java
@@ -15,6 +15,11 @@
 package com.google.androidbrowserhelper.playbilling.digitalgoods;
 
 import android.os.Bundle;
+import android.util.Log;
+
+import com.android.billingclient.api.BillingClient;
+import com.android.billingclient.api.BillingResult;
+import com.google.androidbrowserhelper.playbilling.provider.BillingWrapper;
 
 import androidx.annotation.Nullable;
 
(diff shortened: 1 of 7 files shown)


playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/GetDetailsCall.java
```diff
@@ -60,7 +64,13 @@ public static GetDetailsCall create(@Nullable Bundle args,
     }
 
     /** Calls the callback provided in the constructor with serialized forms of the parameters. */
-    private void respond(BillingResult code, List<SkuDetails> detailsList) {
+    private void respond(BillingResult result, List<SkuDetails> detailsList) {
+        int responseCode = result.getResponseCode();
+        if (responseCode != BillingClient.BillingResponseCode.OK) {
+            Log.w(TAG, "Billing returned non-OK response code: " + responseCode + ", "
+                    + result.getDebugMessage());
+        }
```
PEConn, line 72: I'll probably extract this code out to a different class (and dedupe with `AcknowledgeCall`) in a future PR.

andreban (approved):
LGTM, Thank you

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/AcknowledgeCall.java
```diff
@@ -15,21 +15,28 @@
...
 
 import androidx.annotation.Nullable;
 
 /**
  * A class for parsing Digital Goods API calls from the browser and converting them into a format
- * suitable for calling the Play Biling library.
+ * suitable for calling the Play Billing library.
  */
 public class AcknowledgeCall {
+    private static final String TAG = "TwaBilling";
```
padr31, line 31: NIT: I just noticed this is not a unique tag. I don't think it is important in this case, but sometimes we include the entire package name in the tag.

PEConn: I wanted to use the same TAG everywhere here - it makes it more convenient to just do `adb logcat | grep TwaBilling` and see all the relevant info instead of `adb logcat | grep -e GetDetailsCall -e AcknowledgeCall -e BillingRequestHandler` and it should be obvious from the content of the message which class printed it.

(Also, Android guidelines are that tags shouldn't be more than 24 characters so they don't take up too much of the log line - so including the whole package name is probably not the best.)

PR #167, opened 2020-09-07
Make BillingWrapper#connect take callbacks.

PEConn:
This PR changes how clients of BillingWrapper are notified of connection state changes.

It lays the groundwork for a future PR that automatically connects BillingWrapper.


andreban (approved):
LGTM, Thank you

PR #168, opened 2020-09-07
Create a BillingWrapper that ensures we are connected.

PEConn:
This PR adds the `ConnectedBillingWrapper` that takes a `BillingWrapper` and ensures that connect is called (and completes) before calling any of its methods.

### DIFF
--- a/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/ConnectedBillingWrapper.java
+++ b/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/ConnectedBillingWrapper.java
@@ -1,3 +1,17 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/ConnectedBillingWrapper.java
```diff
@@ -0,0 +1,93 @@
+package com.google.androidbrowserhelper.playbilling.digitalgoods;
```
andreban, line 1: nit: Missing license header.

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/ConnectedBillingWrapper.java
```diff
@@ -0,0 +1,93 @@
...
+ */
+public class ConnectedBillingWrapper implements BillingWrapper {
+    private final BillingWrapper mInner;
+
+    private static final int NOT_CONNECTED = 0;
+    private static final int CONNECTING = 1;
+    private static final int CONNECTED = 2;
+    private int mState = NOT_CONNECTED;
+
+    private interface Callback {
```
andreban, line 28: This is more of an observation: this is the same as the `Runnable` interface. Would it make sense to use it instead (or extend if we want to add extra methods)?

andreban (approved):
LGTM, with the missing license header.

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/ConnectedBillingWrapper.java
PEConn: Oh good spot, thanks!

PR #173, opened 2020-09-13
Create OneSignal for TWA

Vegas10128:
Hello Project Developers. 
Is there a possibility to allowing OneSignal notifications passed through TWA Apps yet?
I have brought this up before but on some other network and cant remember where its been about 4 or so months since i discussed it. 

Can this be possible to have OneSignal for TWA Apps?


andreban:
It seems OneSignal supports web push notifications. If you use that in your web app, it should work https://onesignal.com/webpush. Closing this as this is not an actual pull request.

PR #175, opened 2020-09-14
Protect against null detailsList.

PEConn:
Play Billing may provide a `null` `detailsList` (instead of a non-null, empty one) if no SKUs are found. Protect against this.


andreban (approved):
LGTM, Thanks

padr31 (approved):
LGTM

PR #176, opened 2020-09-14
Remove old Play Billing code.

PEConn:
This PR removes the old `bundles` directory which was an early exploratory implementation of Play Billing support.

Since `bundles` was the only part of the code that required Kotlin support, that was removed as well.


andreban (approved):
LGTM, thanks

padr31 (approved):
LGTM

PR #177, opened 2020-09-14
Remove payment provider demo.

PEConn:
The payment provider demo was useful in developing TWA Play Billing support. However it has now been superseded by the `twa-play-billing` demo.

This original demo no longer works now that the test website has been designed for TWA Play Billing (which doesn't require a custom Payment Method).


andreban (approved):
I think we may need to remove the reference to the demo from `settings.gradle`. LGTM, with that.

padr31 (approved):
LGTM, with Andre's comment about settings.gradle

PR #178, opened 2020-09-14
Add payment proxy to PaymentActivity.

PEConn:
This change stores the app that launched the `PaymentActivity` in the `PaymentActivity`'s Intent. In practice this will be the browser that the user is currently using for the TWA.

Additionally this fixes the Payments instrumentation tests that were broken with the move from `AppCompatActivity` to `Activity` (our instrumentation tests require Android devices so can't be run on GitHub actions).


andreban (approved):
LGTM, Thanks!

padr31 (approved):

PR #185, opened 2020-09-23
Unify logging for Play Billing.

PEConn:
This adds debug logging for pretty much everything under the tag `TwaBilling`.

### DIFF
--- a/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/DigitalGoodsRequestHandler.java
+++ b/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/DigitalGoodsRequestHandler.java
@@ -31,9 +31,7 @@ import androidx.browser.trusted.TrustedWebActivityCallbackRemote;
 
 public class DigitalGoodsRequestHandler implements ExtraCommandHandler {
     private final BillingWrapper mWrapper;
-    private final BillingWrapper.Listener mListener = result -> {
-        throw new IllegalStateException("DigitalGoods does not support Payment Flows.");
-    };
+    private final BillingWrapper.Listener mListener = result -> { };
 
     public DigitalGoodsRequestHandler(Context context) {
         mWrapper = new ConnectedBillingWrapper(BillingWrapperFactory.get(context, mListener));
--- a/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/DigitalGoodsRequestHandler.java
+++ b/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/DigitalGoodsRequestHandler.java
@@ -60,6 +60,8 @@ public class DigitalGoodsRequestHandler implements ExtraCommandHandler {
 
     public boolean handle(@NonNull String commandName, @NonNull Bundle args,
             @Nullable DigitalGoodsCallback callback) {
+        Logging.logCommand(commandName);
+
         switch (commandName) {
             case GetDetailsCall.COMMAND_NAME:
                 GetDetailsCall getDetailsCall = GetDetailsCall.create(args, callback);
--- a/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/Logging.java
+++ b/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/Logging.java
@@ -1,3 +1,17 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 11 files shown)


playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/Logging.java
```diff
@@ -0,0 +1,66 @@
+package com.google.androidbrowserhelper.playbilling.digitalgoods;
```
andreban, line 1: Missing license header.

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/Logging.java
```diff
@@ -0,0 +1,35 @@
+package com.google.androidbrowserhelper.playbilling.provider;
```
andreban, line 1: Missing license header

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/Logging.java
```diff
@@ -0,0 +1,35 @@
...
+import android.util.Log;
+
+import com.android.billingclient.api.BillingClient;
+import com.android.billingclient.api.BillingResult;
+
+/**
+ * Consolidates all the logging for this package in one place.
+ */
+public class Logging {
+    private static final String TAG = "TwaBilling";
```
andreban, line 12: I wonder if it's worth differentiating from the other logging? Eg: `TwaBilling.digitalgoods` and `TwaBilling.provider`.

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/Logging.java
```diff
@@ -0,0 +1,66 @@
...
+
+    static void logGetDetailsResponse(BillingResult result) {
+        int responseCode = result.getResponseCode();
+        Log.d(TAG, "GetDetails returned code " + responseCode);
+        if (responseCode != BillingClient.BillingResponseCode.OK) {
+            Log.d(TAG, result.getDebugMessage());
+        }
+    }
+
+    static void logConnected() {
```
andreban, line 59: Are those two intentionally left empty? Can we add a comment saying it if that's the case?

andreban (changes requested):
Looks good. A couple of missing license headers and a couple of questions.

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/Logging.java
PEConn: Nope, that was an accident.

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/Logging.java
PEConn: I'd like to just do `adb logcat | grep TwaBilling` and get everything. But yeah, I like your idea, I don't want the tags to get too long so I'll do it as `TwaBilling.P` and `TwaBilling.DG`.

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/DigitalGoodsRequestHandler.java
```diff
@@ -31,9 +31,7 @@
 
 public class DigitalGoodsRequestHandler implements ExtraCommandHandler {
     private final BillingWrapper mWrapper;
-    private final BillingWrapper.Listener mListener = result -> {
-        throw new IllegalStateException("DigitalGoods does not support Payment Flows.");
```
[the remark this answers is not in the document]
PEConn: I probably should have made this bit more explicit. But essentially this listener gets triggered by the Play Billing Library whenever someone completes a payment flow. The original reason for having this exception is that the payment flow cannot be triggered through the Digital Goods API. However, apparently the Play Billing Library treats payment flow completion as a global event. This results in the following bug:

1. We use the Digital Goods API to get the price of something (it has its own Play Billing Library connection).
2. We then launch PaymentActivity, which also connects to Play Billing and attaches its own listener.
3. Once the payment flow is complete, both PaymentActivity's listener and DigitalGoodsRequestHandler's listener are triggered.

So basically, if someone used the Digital Goods API before launching the Payment Activity, completing the payment flow would crash the app.

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/Logging.java
PEConn: Done

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/Logging.java
PEConn: Done

andreban (approved):
LGTM, Thanks!

Issue #181, opened 2020-09-18
Consider renaming the "Site Settings" shortcut


andreban:
As an application downloaded from the Play Store, the "Site Settings" may not be clear for users. Consider renaming to "Application Settings" or something similar.

More context:
- https://twitter.com/tomayac/status/1306840186219892736
- https://twitter.com/firt/status/1307014143317364736

@padr31 FYI

FluorescentHallucinogen:
Wouldn't it be better to add the ability to customize the name and icon?

See the rationale: https://twitter.com/daviddalbusco/status/1307014586743361536

FluorescentHallucinogen:
@b1tr0t cc.

andreban:
Adding the image from the tweet @FluorescentHallucinogen mentioned. The icon is looking very off in this image. Another point to consider here is internationalisation, which doesn't seem to be supported in the code.

![EiNyDG9XgAM39ef](https://user-images.githubusercontent.com/1733592/93636024-1b5e0a00-f9eb-11ea-997d-07ac3b5cc35a.jpeg)

peterpeterparker:
Thank you @FluorescentHallucinogen 👍 

As I meant in my tweet and  above screenshot, I would really appreciate the ability to style the icon (not just for design reason but also for accessibility reason) and I also personally would rather like to display "App Settings"  instead of "Site Settings" but, that's a personal point of view of course.

Regarding i18n, definitely would be nice. I think, or if I remember correctly, there is already an ongoing topic regarding the weblinks as currently you can't translate these neither.

gstepniewski-google:
Closing due to no updates in 5+ years (and we're currently working on adding string translations)

peterpeterparker:
Yeah, after 5+ years, I couldn't care less about it anymore. Good call to close this.

PR #186, opened 2020-09-23
Add support for overriding site settings icon

padr31:
Responding to this issue https://github.com/GoogleChrome/android-browser-helper/issues/181, we are adding support for overriding the site settings icon.

When adding the shortcut, we check whether override_ic_site_settings is present as a drawable and use it. Otherwise, we fall back to the default ic_site_settings icon.

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivity.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ManageDataLauncherActivity.java
@@ -86,6 +86,9 @@ public class ManageDataLauncherActivity extends Activity {
     public static final String CATEGORY_LAUNCH_SITE_SETTINGS =
             "androidx.browser.trusted.category.LaunchSiteSettings";
 
+    public static final String OVERRIDE_IC_SITE_SETTINGS_ID =
+            "drawable/override_ic_site_settings";
+
     @Nullable
     private String mProviderPackage;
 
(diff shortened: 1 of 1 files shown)


PEConn (approved):
Code looks good.

I'm a little unsure what the best strategy should be:
1. Have a default icon in ABH and let people override it (this PR).
2. Have no default icon and require the developer to provide one. This would let the developer turn this feature off by omitting and icon and we could include a default icon in bubblewrap.

(Yeah, I know I was the one who suggested approach 1...)

What do you think? (I'm happy landing the code as is and then seeing how people use it if we have no strong opinions.)

padr31:
> Code looks good.
> 
> I'm a little unsure what the best strategy should be:
> 
> 1. Have a default icon in ABH and let people override it (this PR).
> 2. Have no default icon and require the developer to provide one. This would let the developer turn this feature off by omitting and icon and we could include a default icon in bubblewrap.
> 
> (Yeah, I know I was the one who suggested approach 1...)
> 
> What do you think? (I'm happy landing the code as is and then seeing how people use it if we have no strong opinions.)

I think approach 1 makes more sense because:
- Lets people who just want to create a TWA using browser-helper and not bubblewrap to use the default
- It is flexible for others - users of both bubblewrap, and browser-helper only can override the icon

Approach 2 would take away the default from browser-helper-only users.

PEConn (approved):
All LGTM then.

andreban (approved):
LGTM, thank you!

PR #187, opened 2020-09-25
Pass purchase token to browser.

PEConn:
Hey,

There are three parts to this PR:
1. Adding the purchase token given by the Play Billing flow to the Bundle returned to the browser.
2. Changing `BillingWrapper.Listener#onPurchaseFlowComplete` to take the token as a parameter so we can move it around.
3. Changing the logging added in the last PR around a bit.

### DIFF
--- a/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/MockBillingWrapper.java
+++ b/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/MockBillingWrapper.java
@@ -111,7 +111,7 @@ public class MockBillingWrapper implements BillingWrapper {
     }
 
     public void triggerOnPurchasesUpdated() {
-        mListener.onPurchaseFlowComplete(toResult(BillingClient.BillingResponseCode.OK));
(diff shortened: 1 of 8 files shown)


playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/MockBillingWrapper.java
```diff
@@ -111,7 +111,7 @@ public void triggerConsume(int responseCode, String token) {
     }
 
     public void triggerOnPurchasesUpdated() {
-        mListener.onPurchaseFlowComplete(toResult(BillingClient.BillingResponseCode.OK));
+        mListener.onPurchaseFlowComplete(toResult(BillingClient.BillingResponseCode.OK), "");
```
padr31, line 114: Is the empty "" token a temporary placeholder?

PEConn: This is just test code, so I was going to use "" until we have some tests that actually use the value.

andreban (approved):
LGTM, Thank you!

PR #189, opened 2020-10-01
Fix style in alpha branch demos

EiraGe:
These two demos are alpha branch only and their style were not updated
after we removed AppCompat.

@andreban  PTAL Thanks.


andreban (approved):
LGTM, Thanks for catching those two, @EiraGe!

PR #190, opened 2020-10-01
Fix location delegation demo

EiraGe:
The demo does not work on Android 11 because ExtraFeaturesService was not public and causes a IllegalAccessException.
Make it public fix the issue.

@andreban PTAL Thanks


andreban:
Tested only making `ExtraFeaturesService` and its constructor public and it fixes it for me, without moving the service tag around.

EiraGe:
Interesting, it wasn't work yesterday, but it seems work now. Reverted the manifest change

andreban (approved):
LGTM, Thanks!

PR #191, opened 2020-10-04
Added trailing slash to WebView fallback referrer

andreban:
- Ensures compatibility with the Chrome implementation.
- Ensures it differentiates com.example from com.example.subdomain.


PEConn (approved):

PEConn:
I can see your first point (parity with what CCTs do), but can you expand on your second a bit?

andreban:
Sure - Inside a web app, a developer can check the user is coming from the TWA  by looking if the referrer matches `android-app://example.com`. The easiest way being `document.referrer.startsWith('android-app://com.example')` - but this would also match `android-app://com.example.subdomain`. This could be solved with a regex, but appending the trailing slash enables developers to check with `document.referrer.startsWith('android-app://com.example/')`, which is less error prone than a regex and avoids the first problem.

PR #193, opened 2020-10-08
Fixes package for the twa-location-delegation demo

andreban:
- The directory structure didn't match the packageId, so the
  project was unable to find the NotificationDelegationService.
- Updates build.gradle, AndroidManifest and the Java file to match
  the directory structure, as we're standardising in other demos.


PEConn (approved):

mvano (approved):
Thanks!

PR #194, opened 2020-10-12
Report price without currency symbol.

PEConn:
This change reports the price of a SKU without the currency symbol.

It does make the price less pretty (for example, "7.500000" instead of "£7.50"), but the website developer can get back to the pretty version using [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) on their website.


mvano (approved):
Thank you, today I learned about underscores in numeric literals.

andreban (approved):
LGTM, Thanks!

PR #196, opened 2020-10-13
Allow overriding TwaLauncher creation.

PEConn:
In response to [this bug](https://github.com/GoogleChromeLabs/bubblewrap/issues/353).

I'm not sure whether we want to land this because it allows locking TWAs to a specific browser. We need to give some consideration to what happens if that browser is not available.

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
@@ -165,7 +165,7 @@ public class LauncherActivity extends Activity {
 
         addShareDataIfPresent(twaBuilder);
 
-        mTwaLauncher = new TwaLauncher(this);
+        mTwaLauncher = createTwaLauncher();
         mTwaLauncher.launch(twaBuilder,
                 mCustomTabsCallback,
                 mSplashScreenStrategy,
@@ -182,6 +182,10 @@ public class LauncherActivity extends Activity {
 
         ManageDataLauncherActivity.addSiteSettingsShortcut(this,
                 mTwaLauncher.getProviderPackage());
+    }
+
+    protected TwaLauncher createTwaLauncher() {
+        return new TwaLauncher(this);
     }
 
     private boolean splashScreenNeeded() {


andreban (approved):
LGTM

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -184,6 +184,10 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
                 mTwaLauncher.getProviderPackage());
     }
 
+    protected TwaLauncher createTwaLauncher() {
+        return new TwaLauncher(this);
```
mvano, line 188: Given this change, what would happen if the specified browser is not installed?

mvano (commented):
Could you clarify what would happen in the current setup when the specified browser is not installed?

PEConn:
So yeah, this PR itself doesn't change any behaviour, it just allows someone to extend `LauncherActivity` and do:

```
@Override
protected TwaLauncher createTwaLauncher() {
    return new TwaLauncher(this, "my.dream.browser");
}
```

In this case, if `my.dream.browser` doesn't exist, I think we'll crash (although I haven't tried it myself).

We have two options:
1. We can change `TwaLauncher` and make it take a block/allow list of browsers or have some other fancy logic.
2. We leave it up to the person overriding `createTwaLauncher` to care about all that.

This PR goes with option 2 which I think could be good because we don't want developers to hardcode certain browsers. If we went with option 1 we're making that easier for developers.

What do you think?

mvano:
I don't feel strongly for either 1 or 2, so let's follow your preference. It's also the simplest code.

mvano (approved):

h-nibor:
Hi @PEConn! Is there anything holding back this branch from merging, apart from needing to be rebased?

(As a TWA developer, I would like to update our dependency with the present changes integrated.)

h-nibor:
I have a local branch that checks whether a specified browser provides a Custom Tabs service. It implements a fallback in case the browser is not installed or not a valid provider.

Would you like me to open a pull request against this branch?

PEConn:
Hey @h-nibor, I'll get this change landed since on it's own it isn't that opinionated (we still don't want to encourage TWAs being tied to certain browsers).

PEConn:
For the moment you can use [jitpack](https://jitpack.io/) if you want to use gradle to depend on a Android Browser Helper HEAD. As for when we'll be next publishing ABH to Maven, over to @andreban .

h-nibor:
Thanks @PEConn !

PR #201, opened 2020-10-23
Make GetDetails and Purchase work with subscriptions.

PEConn:
Whenever the Play Billing API takes a SkuType as a parameter, we now call that method twice - once with in app purchases and once with subscriptions and combine the results.


andreban (approved):
LGTM, Thanks

PR #204, opened 2020-10-30
Switch consume and acknowledge

nohe427:
Consume makes a purchase able to be repeatedly purchased, whereas acknowledge only charges for the purchase and prevents future purchases of the item.

PAIR=joycetoh8
TESTS=manual, updated unit tests


PEConn (approved):
Thanks for this Alex!

PR #210, opened 2020-11-09
Make BillingResultMerger generic.

PEConn:
This PR makes `BillingResultMerger` generic, which should allow it to be used to merge the results of `queryPurchases` in the near future.

### DIFF
--- a/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/BillingResultMerger.java
+++ b/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/BillingResultMerger.java
@@ -17,8 +17,6 @@ package com.google.androidbrowserhelper.playbilling.digitalgoods;
 import com.android.billingclient.api.BillingClient;
 import com.android.billingclient.api.BillingClient.SkuType;
 import com.android.billingclient.api.BillingResult;
-import com.android.billingclient.api.SkuDetails;
-import com.android.billingclient.api.SkuDetailsResponseListener;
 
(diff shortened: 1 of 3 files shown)


playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/BillingResultMerger.java
```diff
@@ -35,26 +35,30 @@
...
  * we don't need to worry about any concurrency.
  */
-public class BillingResultMerger {
-    private final SkuDetailsResponseListener mOnCombinedResult;
+public class BillingResultMerger<T> {
+    private final ResultListener<T> mOnCombinedResult;
 
     private @Nullable BillingResult mInAppResult;
-    private @Nullable List<SkuDetails> mInAppDetailsList;
+    private @Nullable List<T> mInAppDetailsList;
```
mvano, line 42: Will member names such as this, and arguments elsewhere, also be changed? I'd expect mInAppDetailsList to always be a List<SkuDetails> for example.

mvano (approved):

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/BillingResultMerger.java
PEConn: It is always going to be a list (the type is `List<T>`)...

Although I did change the names to get rid of "Details", so it's now `mInAppResultsList`.

PR #214, opened 2020-11-10
Update Play Billing library.

PEConn:
This PR updates the Play Billing library dependency to version 3.0.1. It also removes an unused Kotlin standard library dependency and updates some tests that were broken by the Play Billing upgrade.


mvano (approved):

andreban (approved):
LGTM, Thank you!

PR #216, opened 2020-11-11
Update keys used in PurchaseDetails.

PEConn:
This PR updates the code to match this change in Chromium:

  https://chromium-review.googlesource.com/c/chromium/src/+/2529550


mvano (approved):

andreban (approved):
LGTM

PR #219, opened 2020-11-13
Add a version change log to MainActivity

PEConn:
This PR should make it easier for testers to know which issues have been fixed (and which version they are using).

This also makes the demo app support Lollipop.


mvano (approved):

andreban (approved):
LGTM

PR #206, opened 2020-11-04
When running on Chrome OS, set ARC_PAYMENT_APP as the last launched provider.

jeevan-shikaram:
This pull requests registers ARC_PAYMENT_APP as the last launched provider when running on Chrome OS, allowing Chrome OS  to query Digital Goods API.

PR #222, opened 2020-11-17
Merge 2 commits that provide Chrome OS support

jeevan-shikaram:
This pull request cherrypicks two commits that add Chrome OS support:
- Chrome OS Play billing support #162
- When running on Chrome OS, set ARC_PAYMENT_APP as the last launched provider. #206

Thanks!


andreban (commented):
If I understood correctly, we're picking some changes from `master` and applying to `alpha`. Is this correct?

jeevan-shikaram:
Yes that's right. Since we are using the `alpha` branch to develop the TWA DGAPI and Play Billing support, we'll need to merge these commits in to test.

PEConn (approved):
LGTM

PR #225, opened 2020-11-18
Set ARC_PAYMENT_APP as a trusted package.

jeevan-shikaram:
Set the ARC_PAYMENT_APP package name to be trusted so that Chrome OS can query the digital goods API.


PEConn (approved):
LGTM

PR #227, opened 2020-11-23
Move Token storing code to where ChromeOS will hit it too.

PEConn:
This PR moves the code for setting the verified provider so that it now executes no matter what the launch mode is.

Previously it only executed when the launch mode was TWA, which mean that it wasn't executed when launched on ChromeOS (which from the point of view of the `TwaLauncher` is launched as a Custom Tab). Now it should be run for ChromeOS as well.


mvano (approved):

andreban (approved):
LGTM, Thanks

PR #272, opened 2021-03-24
Offline-First TWA demo (Iteration 1)

demianrenzulli:
Iteration 1 for the "Offline-FIrst TWA demo". See #273 for details on iterations.

### Steps:

- Run the app on the phone. Note: The first time the app is ran you can't be in airplane mode, because Android Studio would throw an error. Alternatively, to run the first test in "offline-first" state, and .apk file can be generated and installed in the phone (outside of Android Studio).
- The TWA with https://airhorner.com/ is launched.
- Close the app and go to Airplane Mode.
- Open the app: the offline fallback screen is shown.
- Clicking the "Retry" button in offline state, doesn't open the TWA, because the user is offline.
- Go back online.
- Click the "Retry" button.
- The TWA is launched.

### Implementation details:

- `OfflineFirstTWALauncherActivity` extends `com.google.androidbrowserhelper.trusted.LauncherActivity` to prevent the default launch of the TWA, by overriding `shouldLaunchImmediately()` to return `false`.
- In the `onCreate` method, ti checks for connectivity. If the user is offline, it shows an Android offline fallback screen, with a "Retry" button.
- If the user is online, the TWA is launched.

### Observations:

In this first iteration the TWA will always show the native fallback screen when it's opened in offline state. This might be good enough in most cases, but we can explore a second iteration where we either:

- Save a flag in sharedPreferences the first time the TWA is launched, assuming that the SW might be registered at that point, and the web offline fallback page can be shown. This is not 100% accurate, since the user  could have cleaned the storage after the first run, in which case the preference would be in an inconsistent state.
- Check if the SW is registered or the html fallback page is in the cache, if so, always launch the TWA in offline state, instead of showing the Android fallback screen.

cc // @andreban @b1tr0t @aranhacarlos


demos/twa-offline-first/build.gradle
```diff
@@ -0,0 +1,39 @@
+plugins {
```
andreban, line 1: nit: please, add license headers to the files and ensure all files finish with an empty line.

demos/twa-offline-first/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
+}
+
+dependencies {
+
+    implementation 'androidx.appcompat:appcompat:1.1.0'
+    implementation 'com.google.android.material:material:1.1.0'
```
andreban, line 33: How much does using this add to the APK size? If the Activity is relatively simple from a UI perspective, we may get  away without adding many extra libraries.

demos/twa-offline-first/src/main/java/com/google/androidbrowserhelper/demos/twa_offline_first/OfflineFirstTWALauncherActivity.java
```diff
@@ -0,0 +1,48 @@
...
+        this.displayScreen();
+    }
+
+    @Override
+    protected boolean shouldLaunchImmediately() {
+        return false;
+    }
+
+    private void displayScreen() {
+        if(isOnline()) {
```
andreban, line 25: style: space between `if` and the parenthesis.
```suggestion
        if (isOnline()) {
```

demos/twa-offline-first/src/main/java/com/google/androidbrowserhelper/demos/twa_offline_first/OfflineFirstTWALauncherActivity.java
```diff
@@ -0,0 +1,48 @@
...
+        super.onCreate(savedInstanceState);
+        this.displayScreen();
+    }
+
+    @Override
+    protected boolean shouldLaunchImmediately() {
+        return false;
+    }
+
+    private void displayScreen() {
```
andreban, line 24: I got confused by the method name when reading the method below. What about renaming it to something like `tryLaunchTwa()`?
```suggestion
    private void tryLaunchTwa() {
```

demos/twa-offline-first/src/main/java/com/google/androidbrowserhelper/demos/twa_offline_first/OfflineFirstTWALauncherActivity.java
```diff
@@ -0,0 +1,48 @@
...
+        if(isOnline()) {
+            launchTwa();
+        } else {
+            launchCustomOfflineScreen();
+        }
+    }
+
+    private void launchCustomOfflineScreen() {
```
andreban, line 32: nit: Launch makes me thing we're launching an Activity, but we're not actually launching anything here, and just setting the content of the currentActivity. Maybe rename this to `renderOfflineFallback()` or something similar?
```suggestion
    private void renderOfflineFallback() {
```

andreban (commented):
Dropped some notes.

demianrenzulli:
@andreban thanks for your review. I think I worked on all your comments now.

Regarding dependencies, I cleaned up most of them. I left  `com.google.android.material:material:1.1.0`, which lets me use the `android:theme="@style/Theme.Design.NoActionBar"` theme in `AndroidManfiest.xml`, to hide the title bar in the offline fallback screen, and make it look more similar to airhorner.

That seems to be the easiest way to  do it. You can also do it programmatically, but that would add noise to the Activity code.

 Let me know your thoughts.

demos/twa-offline-first/build.gradle
```diff
@@ -0,0 +1,50 @@
+/*
+ *    Copyright 2019 Google LLC
```
PEConn, line 2: Nit: could you go through all of the copyright headers and update the year to 2021 please?

settings.gradle
```diff
@@ -17,3 +17,4 @@ include ':demos:twa-orientation'
 include ':demos:twa-play-billing'
 include ':demos:twa-web-share-target'
 include ':demos:twa-webview-fallback'
+include ':demos:twa-offline-first'
```
PEConn, line 20: nit: Could you keep the demos sorted in alphabetical order?

demos/twa-offline-first/src/main/java/com/google/androidbrowserhelper/demos/twa_offline_first/OfflineFirstTWALauncherActivity.java
```diff
@@ -0,0 +1,62 @@
...
+import android.os.Bundle;
+import android.view.View;
+import android.widget.Button;
+
+public class OfflineFirstTWALauncherActivity extends LauncherActivity {
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        this.tryLaunchTwa();
```
PEConn, line 30: nit: no need for `this`.

demos/twa-offline-first/src/main/java/com/google/androidbrowserhelper/demos/twa_offline_first/OfflineFirstTWALauncherActivity.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    protected boolean shouldLaunchImmediately() {
+        return false;
```
PEConn, line 35: Could you add a comment just explaining the consequences of this line? (This is a demo so we may as well make things as obvious as possible.)

demos/twa-offline-first/src/main/java/com/google/androidbrowserhelper/demos/twa_offline_first/OfflineFirstTWALauncherActivity.java
```diff
@@ -0,0 +1,62 @@
...
+            launchTwa();
+        } else {
+            renderOfflineFallback();
+        }
+    }
+
+    private void renderOfflineFallback() {
+        setContentView(R.layout.activity_offline_first_twa);
+
+        final Button retryBtn = this.findViewById(R.id.retry_btn);
```
PEConn, line 49: nit: We don't normally put final on local variables.

demos/twa-offline-first/src/main/java/com/google/androidbrowserhelper/demos/twa_offline_first/OfflineFirstTWALauncherActivity.java
```diff
@@ -0,0 +1,62 @@
...
+        retryBtn.setOnClickListener(new View.OnClickListener() {
+            public void onClick(View v) {
+                tryLaunchTwa();
```
PEConn, line 52: If the user is offline, and they click the button when they're still offline, we're going to be repeating all of this code even though there's no need to.

Maybe tryLaunchTwa should be something like:

```
if (isOnline()) {
  launchTwa();
} else if (!mOfflineFallbackRendered) {
  mOfflineFallbackRendered = true;
  renderOfflineFallback();
}
```

Or alternatively we just change this line to:

```
if (isOnline()) launchTwa();
```

Which is probably a bit simpler.

PEConn (commented):
Hey, thanks for the contribution.

I've left a few comments - they're all pretty simple things.

demianrenzulli:
Thanks for taking the time to review this PR @PEConn! I think I went through all your comments.

Let me know if you  have any additional thoughts.

Cheers.

PEConn (approved):

andreban (approved):
LGTM too

Issue #273, opened 2021-03-24
Add na offline-first-run Trusted Web Activity demo


andreban:
When a PWA using Trusted Web Activity is installed from an application store, the Service Worker and therefore its offline capabilities won't be installed until it runs for the first time.

This creates a risk where, if the user is offline when they run the app for the first time, they will see the Chrome Dinosaur instead of an application specific offline page.

A workaround for this issue is to check for connectivity when the application runs for the first time and show a custom Activity in that case. When the web application is launched successfully for the first time, that information is stored and future offline launches can use the web app offline capability.

**[ ] Iteration 1:** The Android application checks for connectivity before launching the web app and launches a custom offline activity if connection is not available.
**[ ] Iteration 2:** Information on the web application being launched successfully is stored in the shared storage and future offline launches will use the web app instead of the Android falllback.

demianrenzulli:
Adding the related bubblewrap implementation that we can use as inspiration for iteration 2:

https://github.com/GoogleChromeLabs/bubblewrap/blob/main/packages/core/src/lib/features/FirstRunFlagFeature.ts

demianrenzulli:
As discussed, we'll first work in getting the iteration 1 approved and uploaded to the repo, and then we'll work on iteration 2 on top of it.

@andreban: question about iteration 2: Taking a look at [the helper class](https://github.com/GoogleChromeLabs/bubblewrap/blob/main/packages/core/src/lib/features/FirstRunFlagFeature.ts) you've shared, I don't see exactly how the following things are being checked:

- How do we know if the service worker registration succeeded in the first run and the fallback page is available in the cache.
- (Related to the previous point): how can we make sure that, after a successful first run, the user hasn't cleaned up the browser storage, and therefore, removed the SW and offline page for the TWA?

I assume that for that functionality to be made, we would need to communicate Android code (from the demo), with web code (to check for  SW registration and caches), and that's not possible.

Thoughts are welcome.

andreban:
To my knowledge, there's no way to know that the Service Worker was successfully installed from the host app, so we just trust that `launchTwa()` launched **something** - Either the TWA or the fallback approach.

 @PEConn in case he has any ideas here.

demianrenzulli:
[...]

PR #275, opened 2021-03-30
Offline-First TWA demo: Iteration 2

demianrenzulli:
As indicated in #273, **iteration 2** stores a boolean in `SharedPreferences` after the first call to l`aunchTwa()`. Subsequent runs will check for the existence of that value to decide if launching the TWA directly or checking for network state first.

This iteration doesn't perform any additional checks (for example, if the TWA actually launched successfully or not). We might want to work on that in a third iteration using [DelegationService](https://github.com/GoogleChrome/android-browser-helper/blob/main/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java), in case that's possible.

**Note:** Testing this one is a bit more tricky, since you actually  need to be in offline state in the first run, and Android Studio breaks when moving to airplane mode. 
I could perform the tests by  generating an .apk and installing it in my device (as indicated [here](https://stackoverflow.com/questions/16709848/build-unsigned-apk-file-with-android-studio)).

cc // @andreban


demos/twa-offline-first/README.md
```diff
@@ -0,0 +1,5 @@
+# Offline First TWA Demo
+
+This demo shows how to build a TWA that shows an offline screen when there's no connectivity, even the first time the user opens the app (whe the the service worker hasn't been registered yet).
```
PEConn, line 3: nit: s/whe/when/

demos/twa-offline-first/src/main/java/com/google/androidbrowserhelper/demos/twa_offline_first/OfflineFirstTWALauncherActivity.java
```diff
@@ -38,27 +40,47 @@ protected boolean shouldLaunchImmediately() {
...
+        } else if (isOnline()) {
+            firstTimeLaunchTwa();
         } else {
             renderOfflineFallback();
         }
     }
 
+    private boolean hasTwaLaunchedSuccessfully() {
+        // Return true if the preference "twa_launched_successfully" has already been set.
+        SharedPreferences sharedPref = getSharedPreferences(getString(R.string.twa_offline_first_preferences_file_key), Context.MODE_PRIVATE);
```
PEConn, line 57: This may cause a StrictMode violation. Generally the first time you access SharedPreferences it does a disk read, but then for subsequent times it's fine. There's not really much we can do about it though - reading that value from the file is on our critical path, so we need to do it.

In order to stop the demo app from complaining about this on people's phones, we may need to surround this line with:

```
StrictMode.ThreadPolicy oldPolicy = StrictMode.getThreadPolicy();
try {
  StrictMode.setThreadPolicy(StrictMode.ThreadPolicy.LAX);
  
  // Do the disk read here.
} finally {
  StrictMode.setThreadPolicy(oldPolicy);
}
```

demianrenzulli:
Thanks for your comments @PEConn!

I ended up adding the additional code surrounding the first disk read operation to make sure it doesn't break in developer's phones.
I added comments explaining it, I hope I did it well.

I tested the app again and it works as expected.

PEConn (approved):
This LGTM, over to @andreban

andreban (approved):
LGTM, Thanks!

PR #280, opened 2021-04-12
Improves the Custom tabs / Headers demo

andreban:
- Adds a default list of browsers when calling `getPackageName()`
otherwise the method only returns a browser if a default is set
and it supports Custom Tabs.
- Modify `onStop()` to only call `unbindService()` if `mSession`
is not null. This is because `mConnection` is always non-null, as
it is set unconditionally on `onStart()`.
- Adds a README.md file with pointers to blogpost and instructions.
- Adds extra comments with instructions to the code.

### DIFF
--- a/demos/custom-tabs-headers/README.md
+++ b/demos/custom-tabs-headers/README.md
@@ -1,0 +1,19 @@
+Custom Tabs Custom Headers Demo
+===============================
+
+This demo shows how to add custom headers to a web request when opening an URL with Custom Tabs.
+
+Adding custom headers only works when the developer owns both the application using Custom Tabs and
+the origin being opened, and can prove ownership via [Digital Asset Links][2].
+
+Check out [this blogpost][1] for more details on how the implementation works and steps to setup
+the Digital Asset Links validation.
+
+:bangbang: **Important:** This project is using a demo page hosted at Glitch. However, the demo will
+only work if the Digital Asset Links validation is successful. So, when testing on your computer,
+remix the project at https://glitch.com/edit/#!/custom-tabs-custom-he, edit the file under
+`public/.well-known/assetlinks.json` with your own SHA-256 fingerprint (use Tools > Terminal
+to find and edit the file), and update the opened by the application.
+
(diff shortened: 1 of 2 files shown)


demos/custom-tabs-headers/README.md
```diff
@@ -0,0 +1,19 @@
...
+the origin being opened, and can prove ownership via [Digital Asset Links][2].
+
+Check out [this blogpost][1] for more details on how the implementation works and steps to setup
+the Digital Asset Links validation.
+
+:bangbang: **Important:** This project is using a demo page hosted at Glitch. However, the demo will
+only work if the Digital Asset Links validation is successful. So, when testing on your computer,
+remix the project at https://glitch.com/edit/#!/custom-tabs-custom-he, edit the file under
+`public/.well-known/assetlinks.json` with your own SHA-256 fingerprint (use Tools > Terminal
+to find and edit the file), and update the opened by the application.
```
PEConn, line 16: nit: .. update the *URL* opened by ...

demos/custom-tabs-headers/src/main/java/com/google/androidbrowserhelper/demos/customtabsheaders/MainActivity.java
```diff
@@ -74,10 +81,18 @@ public void onServiceDisconnected(ComponentName componentName) {
             }
         };
 
-        String packageName = CustomTabsClient.getPackageName(MainActivity.this, null);
+        //Add package names for other browsers that support Custom Tabs and custom headers.
+        List<String> packageNames = Arrays.asList(
```
PEConn, line 85: Could you please:

1. Add `com.google.android.apps.chrome` (local builds of Chrome) to the list.
2. Reversing the list order, so we go from least stable to most. (Generally if someone has a local build of Chrome installed, they want their apps to use that.)

PEConn (approved):
LGTM with some small comments.

demos/custom-tabs-headers/README.md
andreban: done!

PR #284, opened 2021-04-28
Prevent relaunching the TWA on subsequent Launcher launches.

PEConn:
I believe this should fix https://github.com/GoogleChromeLabs/bubblewrap/issues/469 .

See comment in LauncherActivity#onCreate for more info.


andreban (approved):
LGTM!

PR #290, opened 2021-05-11
Don't finish TWAs started with Share Intents.

PEConn:
This should fix [this issue](https://github.com/GoogleChromeLabs/bubblewrap/issues/524).


andreban (approved):
LGTM!

PR #294, opened 2021-06-02
Bumps version for abh and locationdelegation

andreban:
- abh incorporates changes on the launcher activity.
- locationdelegation removes the dependency on
  androidx.appcompat.


PEConn (approved):
LGTM

PR #296, opened 2021-06-04
Fixes build for the location delegation demo

andreban:
- Uses debugImplementation and releaseImplementation and makes
build.gradle compatible with the libraries.


PEConn (approved):
Thanks :-)

PR #297, opened 2021-06-07
Fixes asynch launchTwa freezing splash screens

### LINKED ISSUE #281: Overriding shouldLaunchImmediately() on the first TWA's launch could have an unexpected behavior
Hi, here are the complete details of an issue/bug concerning the override of the method shouldLaunchImmediately().
This issue could be a bug but I'm not sure.
I tried to strictly refer to the available documentation but I'm not sure where the error comes from.

**The goal to be achieved**
The purpose of the code below is to override the **shouldLaunchImmediately()** method in a class that extends the android-browser-helper's **LauncherActivity** to add a query string to the underlying web app url.
The need is also to render the TWA only after generating a new **Firebase Token** which is retrieved through the **addOnCompleteListener** event handler.

The steps are the following:
**1)** If this is the first time app is launched, I generate the firebase token in the following manner:
FirebaseMessaging.getInstance().GetToken().AddOnCompleteListener(task -> {....})

**2)** Once the token has been recovered, I interface via JsonObjectRequest to an api on a dedicated server. The Guid returned by the API's http POST request is assigned to a local variable declared in the class extending LauncherActivity.

**3)** Overriding **getLaunchingUrl()** checks that the Guid has been assigned to the local variable **fcm_my_server_installId**.

**4)** If the variable fcm_my_server_installId has been set, the Guid is added to the url as a querystring in the following form:
https://myapp/home?installationId=guid-value
 
**5)** If the variable has not been set, this is not the first launch of the app. In that case, **shouldLaunchImmediately()** method returns "true" and **getLaunchingUrl()** returns the url with no query string.

**Describe the bug**
- Sometimes the following code:
String token = task.getResult();
throws com.google.android.gms.tasks.RuntimeExecutionException exception (Service Unavailable), but not always.
 (With the Pixel 3 XL virtual device it happened).
- App splash screen gets stuck on screen indefinitely (Always happens)
- Starting from the second loading (without query string) the app loads correctly (It always happens).
- If (keeping the same code) I only remove the override of **shouldLaunchImmediately()** method by restoring the default behavior of the TWA, the app is generally loaded correctly both the first time and the following ones, but the method overriding **getLaunchingUrl()** is launched twice and two instances of the app are loaded on the virtual device screen.

 
**Code Snippets**

```java
package ...;
import ...

public class CustomTwaActivity extends LauncherActivity {

    private String fcm_my_server_installId;

    @Override

	    boolean bGetNewToken = check_If_Get_New_FCMToken();
        if (bGetNewToken) {
            set_Firebase_New_Token();
        }	
    }

    @Override
    protected boolean shouldLaunchImmediately() {
        // if shouldLaunchImmediately() returns 'false' we want to wait until we got a new Firebase token
        // and then launch the Trusted Web Activity with 'launchTwa()'.
        boolean bGetNewToken = check_If_Get_New_FCMToken();
        return !bGetNewToken;
    }

    private boolean check_If_Get_New_FCMToken()
    {
      return (Is It the first load?);
    }

    private void set_Firebase_New_Token() {
        try {
            FirebaseMessaging firebaseMessaging = FirebaseMessaging.getInstance();
            // Start the asynchronous task to get a new Firebase Token.
            firebaseMessaging.getToken().addOnCompleteListener(task -> {
                try {					
                    // Once the task is complete, use token to get my server installId 
                    // and use it in the method getLaunchingUrl().
                    String token = task.getResult();
                    set_local_fcm_installId_from_my_server(getApplicationContext(), token);
					
                } catch (com.google.android.gms.tasks.RuntimeExecutionException ex) {
                    launchTwa();
                }
            });
        } catch (Exception e) {
            Log.e(TAG, getString(R.string.FailToReg_FCMtoken), e);
        }
    }

    private void set_local_fcm_installId_from_my_server(Context context, String token) {
        try {
            JSONObject jsonRequest = new JSONObject();
            jsonRequest.put("inputFieldName", token);
            RequestQueue queue = Volley.newRequestQueue(context);

            JsonObjectRequest jsonObjectRequest = new JsonObjectRequest
                    (Request.Method.POST, "https://my-remote-api", jsonRequest,
                            response -> {
                                try {
                                    if (response == null) { return; }
                                    fcm_my_server_installId = response.getString("outputFieldName")
                                } finally {
                                    queue.stop();
                                    launchTwa();
                                }
                            },
                            error -> {
                                Throwable cause = error.getCause();
                                String msg = error.getLocalizedMessage();
                                if (cause != null) {
                                    msg = (msg == null) ? "Unknown error" : msg;
                                    Log.println(Log.ERROR, cause.getMessage(), msg);
                                }
                                queue.stop();
                            }
                    );
            // Add the request to the RequestQueue.
            queue.add(jsonObjectRequest);
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
	
    @Override
    protected Uri getLaunchingUrl() {
        Uri uri = super.getLaunchingUrl();
        Uri resultUri;
        if (fcm_my_server_installId != null) {
            resultUri = uri.buildUpon()
                    .appendQueryParameter("installationId", fcm_my_server_installId)
                    .build();
        } else {
            resultUri = uri;
        }
        return resultUri;
    }

}

```

**To Reproduce**
Launch the code through one of the default virtual devices of Visual Studio Code.

**Expected behavior**
Twa is correctly loaded both the first time (with the query string set) and the following ones (without any query string set)

**Smartphones**
 - Device: All the default virtual devices installed in Visual Studio Code (Pixel 3 XL, 4.7WXGA, Galaxy Nexus...)
 - OS: Android 10
 - Browsers Installed: Chrome
 - Browser Versions: 83
 - android-browser-helper library version: 2.2.0

andreban:
- Moved code that builds the splash screen strategy to run on
  `onCreate()`, so they will already be available when
  `onEnterAnimationComplete()` is invoked.
- When `finish()` is called early, on `onCreate()`, to bring an
  existing Activity to front, `launchTwa()` shouldn't be invoked.
- We update the Firebase demo to reflect that and add a guard in
  `launchTwa()`.

Closes #281

### DIFF
--- a/demos/twa-firebase-analytics/src/main/java/com/google/androidbrowserhelper/demos/twa_firebase_analytics/FirebaseAnalyticsLauncherActivity.java
+++ b/demos/twa-firebase-analytics/src/main/java/com/google/androidbrowserhelper/demos/twa_firebase_analytics/FirebaseAnalyticsLauncherActivity.java
@@ -36,8 +36,7 @@ public class FirebaseAnalyticsLauncherActivity extends LauncherActivity {
     protected void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
-        // The Activity was finished before the launch because an existing instance of
-        // the app was brought to front. We don't want to call launchTwa() in this case.
+        // `super.onCreate()` may have called `finish()`. In this case, we don't do any work.
         if (isFinishing()) {
             return;
         }
--- a/demos/twa-firebase-analytics/src/main/java/com/google/androidbrowserhelper/demos/twa_firebase_analytics/FirebaseAnalyticsLauncherActivity.java
+++ b/demos/twa-firebase-analytics/src/main/java/com/google/androidbrowserhelper/demos/twa_firebase_analytics/FirebaseAnalyticsLauncherActivity.java
@@ -46,10 +46,10 @@ public class FirebaseAnalyticsLauncherActivity extends LauncherActivity {
 
         // Start the asynchronous task to get the Firebase application instance id.
         firebaseAnalytics.getAppInstanceId().addOnCompleteListener(task -> {
-            // Once the task is complete, save the instance id so it can be used by
-            // getLaunchingUrl().
-            mAppInstanceId = task.getResult();
-            launchTwa();
+                // Once the task is complete, save the instance id so it can be used by
+                // getLaunchingUrl().
+                mAppInstanceId = task.getResult();
+                launchTwa();
         });
     }
 
--- a/demos/twa-firebase-analytics/src/main/java/com/google/androidbrowserhelper/demos/twa_firebase_analytics/FirebaseAnalyticsLauncherActivity.java
+++ b/demos/twa-firebase-analytics/src/main/java/com/google/androidbrowserhelper/demos/twa_firebase_analytics/FirebaseAnalyticsLauncherActivity.java
@@ -36,15 +36,19 @@ public class FirebaseAnalyticsLauncherActivity extends LauncherActivity {
     protected void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
-        FirebaseAnalytics firebaseAnalytics = FirebaseAnalytics.getInstance(this);
+        // The Activity was finished before the launch because an existing instance of
+        // the app was brought to front. We don't want to call launchTwa() in this case.
+        if (!isFinishing()) {
+            FirebaseAnalytics firebaseAnalytics = FirebaseAnalytics.getInstance(this);
 
-        // Start the asynchronous task to get the Firebase application instance id.
-        firebaseAnalytics.getAppInstanceId().addOnCompleteListener(task -> {
+            // Start the asynchronous task to get the Firebase application instance id.
+            firebaseAnalytics.getAppInstanceId().addOnCompleteListener(task -> {
                 // Once the task is complete, save the instance id so it can be used by
                 // getLaunchingUrl().
                 mAppInstanceId = task.getResult();
                 launchTwa();
-        });
+            });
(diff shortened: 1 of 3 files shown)


demos/twa-firebase-analytics/src/main/java/com/google/androidbrowserhelper/demos/twa_firebase_analytics/FirebaseAnalyticsLauncherActivity.java
```diff
@@ -36,6 +36,12 @@
     protected void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
+        // The Activity was finished before the launch because an existing instance of
+        // the app was brought to front. We don't want to call launchTwa() in this case.
```
PEConn, line 40: I find this comment a little confusing - would you be happy with:

super.onCreate may have called finish, in this case, don't do any work.

PEConn (approved):
LGTM with comment nit.

demos/twa-firebase-analytics/src/main/java/com/google/androidbrowserhelper/demos/twa_firebase_analytics/FirebaseAnalyticsLauncherActivity.java
andreban: done.

paulsharratt0:
You guys work fast, thanks for fixing!

paulsharratt0:
Can we bump the version up to include this fix? Thanks!

andreban:
This has been published on [`2.2.2`](https://maven.google.com/web/index.html?q=#com.google.androidbrowserhelper:androidbrowserhelper:2.2.2).

PR #302, opened 2021-06-08
Remove IMMEDIATE_WITHOUT_PRORATION mode

joycetoh8:


PEConn (commented):
The code looks good, I just wanted to check that we've thought through the resulting behaviour.

If the developer inputs `immediateWithoutProration` we will ignore it - treating the request as if no proration mode had been set at all. Alternatively, I guess we could fail the Payment Request (this would require a slightly bigger code change).

I'll mention this on the chat.

PEConn (approved):
This is good, thanks.

Once we have some public documentation about this, it would be good to add a link to it - maybe both in the code here and in the error message for the developer.

joycetoh8:
I have a pending PR in chromeos.dev to add a note about why we've temporarily disabled `immediateWithoutProration`. And updated this PR to also have a link to that page. I don't know if any other documentation that needs to be updated. Andre's page at https://developer.chrome.com/docs/android/trusted-web-activity/receive-payments-play-billing/ doesn't mention proration modes. 

@PEConn I think I need a maintainer to approve running the workflow for this PR. Thanks!

PR #308, opened 2021-06-16
Add missing IMMEDIATE_AND_CHARGE_FULL_PRICE proration mode

joycetoh8:


PEConn (approved):
Whoops, didn't realise I'd missed one out.

Thanks for fixing it Joyce!

joycetoh8:
No problem! A lot of the proration mode names look the same. Thanks for taking a look at the PR! Do you know why the checks are failing and how I should fix it?

PEConn:
It's trying to build your change and cannot find `IMMEDIATE_AND_CHARGE_FULL_PRICE`. It looks like that value was only added [last month in version 4.0.0](https://developer.android.com/google/play/billing/release-notes#4-0) (that also explains why I hadn't added it!).

I think you'll need to update the build.gradle to use the newest version of the billing client [here](https://github.com/GoogleChrome/android-browser-helper/blob/main/playbilling/build.gradle#L53).

joycetoh8:
Ahh good catch, Peter! Makes a lot more sense why you hadn't added it. Cause it didn't exist! haha.

I updated the PBL to 4.0 and had to update some methods that were removed as a result. Do you think the testing and checks coverage is enough to ensure nothing else will be broken?

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/PlayBillingWrapper.java
```diff
@@ -110,18 +110,21 @@ public void consume(String token, ConsumeResponseListener callback) {
...
 
         if (methodData.prorationMode != null) {
-            builder.setReplaceSkusProrationMode(methodData.prorationMode);
+            subUpdateParamsBuilder.setReplaceSkusProrationMode(methodData.prorationMode);
         }
 
+        BillingFlowParams.Builder builder = BillingFlowParams.newBuilder();
+
+        builder.setSkuDetails(sku);
+        builder.setSubscriptionUpdateParams(subUpdateParamsBuilder.build());
```
PEConn, line 126: Do you mind rejigging this code so that we don't call `setSubscriptionUpdateParams` unless the old purchase token or proration mode is set?

joycetoh8: done! Sorry for the delay.

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/PlayBillingWrapper.java
```diff
@@ -110,16 +110,20 @@ public void consume(String token, ConsumeResponseListener callback) {
 
     @Override
     public boolean launchPaymentFlow(Activity activity, SkuDetails sku, MethodData methodData) {
+        BillingFlowParams.SubscriptionUpdateParams.Builder subUpdateParamsBuilder = BillingFlowParams.SubscriptionUpdateParams.newBuilder();
```
PEConn, line 113: nit: Could you wrap this line to 100 characters please?

PEConn (approved):
LGTM % nit

PR #310, opened 2021-06-22
[LocationDelegation] Handle only granted approximate location on S+

EiraGe:
For simplicity, we always request both fine location and coarse location.
When the location prompt show and users only allow the approximate location, treat it as "Allowed"

### DIFF
--- a/locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/PermissionRequestActivity.java
+++ b/locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/PermissionRequestActivity.java
@@ -27,6 +27,8 @@ import android.os.Messenger;
 import android.os.RemoteException;
 import android.text.TextUtils;
 
+import java.util.Arrays;
+
 import androidx.browser.trusted.TrustedWebActivityCallbackRemote;
 import androidx.core.app.ActivityCompat;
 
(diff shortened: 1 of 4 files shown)


EiraGe:
@PEConn @andreban PTAL Thanks

andreban (approved):
Code LGTM. Is this behaviour consistent to what developers expect when running in the browser with regular web permissions?

locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/PermissionRequestActivity.java
```diff
@@ -42,7 +44,8 @@
     private static final String EXTRA_PERMISSIONS = "EXTRA_PERMISSIONS";
     private static final String EXTRA_GRANT_RESULTS = "EXTRA_GRANTED_RESULT";
 
-    private static final String LOCATION_PERMISSION = Manifest.permission.ACCESS_FINE_LOCATION;
+    private static final String[] LOCATION_PERMISSION =
```
PEConn, line 47: Can we make this into a List instead of storing it as an array and converting it into a list when we use it?

PEConn (approved):
This looks good, thanks!

EiraGe:
> Code LGTM. Is this behaviour consistent to what developers expect when running in the browser with regular web permissions?

Yes, this is the same behavior as the web permissions.

locationdelegation/src/main/java/com/google/androidbrowserhelper/locationdelegation/PermissionRequestActivity.java
EiraGe: This is also used in requesting permission, which takes an array as input. I think converting array to list is easier than convert list to array?

PEConn: Ah, fair enough.

PEConn:
LGTM

EiraGe:
@andreban do you mind merging this? Thanks :-)

Issue #312, opened 2021-06-29
SharedPreferencesTokenStore leaks context


theobch:
**Describe the bug**
The context passed to instantiate the [TwaLauncher](https://github.com/GoogleChrome/android-browser-helper/blob/bfe0254d5c3c90c515da4d7b0842194f97715a8a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java#L118) is also used to build the [TokenStore](https://github.com/GoogleChrome/android-browser-helper/blob/bfe0254d5c3c90c515da4d7b0842194f97715a8a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java#L120) object, but this context is never freed, leading to a memory leak when calling activity is destroyed. This is the same type of leak as #283, just one step below.

**To Reproduce**
1. From an Activity use `TwaLauncher#launch()`
2. From the web page, redirect to the Activity through deeplink
3. Handle the Activity result and `finish()` then `twaLauncherInstance.destroy()` in `onDestroy()`
4. The Activity is destroyed but context is still retained

**Expected behavior**

Context should be freed, and this can achieve in two ways:
- Use `context.applicationContext` in [`SharedPreferencesTokenStore` constructor](https://github.com/GoogleChrome/android-browser-helper/blob/bfe0254d5c3c90c515da4d7b0842194f97715a8a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/SharedPreferencesTokenStore.java#L45)
- Add a `clean()` (arbitrary name) method inside `SharedPreferencesTokenStore` that nullify `SharedPreferencesTokenStore.context`

**LeakCanary trace**
```
┬───
│ GC Root: Global variable in native code
│
├─ androidx.browser.customtabs.CustomTabsClient$2 instance
│    Leaking: UNKNOWN
│    Retaining 58.3 kB in 957 objects
│    Anonymous subclass of android.support.customtabs.ICustomTabsCallback$Stub
│    ↓ CustomTabsClient$2.this$0
│                         ~~~~~~
├─ androidx.browser.customtabs.CustomTabsServiceConnection$1 instance
│    Leaking: UNKNOWN
│    Retaining 57.8 kB in 955 objects
│    Anonymous subclass of androidx.browser.customtabs.CustomTabsClient
│    mApplicationContext instance of fr.leboncoin.LBCApplicationDebug
[...]

PR #313, opened 2021-06-29
[SharedPreferencesTokenStore] Use applicationContext

theobch:
Context used to build `SharedPreferencesTokenStore` is never freed, thus
leading to a memory leak when activity which calls `TwaLauncher` is
destroy (`SharedPreferencesTokenStore` is build through
`TwaLauncher` constructor). See #312 for more explanation / trace.

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/SharedPreferencesTokenStore.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/SharedPreferencesTokenStore.java
@@ -43,7 +43,7 @@ public class SharedPreferencesTokenStore implements TokenStore {
      * @param mContext The {@link Context} where the {@link SharedPreferences} will be stored.
      */
     public SharedPreferencesTokenStore(Context mContext) {
-        this.mContext = mContext;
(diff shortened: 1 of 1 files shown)


theobch:
@googlebot I signed it!

theobch:
@googlebot I fixed it.

PEConn:
Could we not fix this by saving the application context instead? So the constructor becomes:

```
    public SharedPreferencesTokenStore(Context context) {
        mContext = context.getApplicationContext();
    }
```

theobch:
@PEConn yes totally ! That's the other solution I proposed in #312 , I've added the `clean()` method for future proofing (i.e, if something else in future needs to be clean when `SharedPreferencesTokenStore` is destroyed), but if `context.applicationContext` is preferred, I can go this way too, no problem 😉

PEConn:
Ah sorry, I was a bit behind on the Context.

I'd vote to use the application context - we can add a cleanup method later if we have other reasons for it.

theobch:
@PEConn  done !

andreban (approved):
LGTM, Thanks for catching this and making a PR with the fix!

PEConn (approved):
Looks good, thanks!

Issue #316, opened 2021-07-07
Support subscription price changes in the Play Billing library.


PEConn:
Allow apps to launch the [Price Change Confirmation Flow](https://developer.android.com/google/play/billing/subscriptions#price-change-launch) by adding a flag in the payment request data, eg:

```
    const supportedInstruments = [{
        supportedMethods: 'https://play.google.com/billing',
        data: {
            sku: "my.sku.here",
            priceChangeConfirmation: "true"
        }
    }]
```

The code change itself will be broken down into the following parts:

- [x] Allow this field to be read into MethodData.
- [x] Create a new PaymentResult type to return.
- [x] Add launchPriceChangeConfirmationFlow to the BillingWrapper and subclasses.
- [x] Wire everything up.
- [x] Add instrumentation tests.

PEConn:
All the code changes have landed, over to @andreban to put it in the next release.

andreban:
Do we need a core release too? Or just billing?

PEConn:
I've not touched anything in core, we just need a billing release.

andreban:
Cool - release is ready, just waiting for approval.

PEConn:
Approved (sorry, I was having some trouble with authenticating yesterday).

rsolomakhin:
Is this OK to close as "fixed"?

PEConn:
Sure.

PR #317, opened 2021-07-07
Read priceChangeConfirmation into MethodData.

PEConn:
This is the first small change for Issue #316.


andreban (approved):
LGTM, Thank you

PR #320, opened 2021-07-13
Fix line wrapping.

PEConn:


andreban (approved):
LGTM - what are we using as the maximum line size? 80 or 100?

PEConn:
I've been using 100 from the [AOSP Java Code guide](https://source.android.com/setup/contribute/code-style#limit-line-length) (the [Chromium Java style guide](https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/java/java.md) is based on that).

PR #330, opened 2021-09-03
Create android2.yml

angie1148:


PEConn:
Hey,

Could you please add a description with this change?

Also, can you confirm that this works? I don't think the dependencies are correct.

AliMousai (approved):
AliMousai

PEConn:
I don't know what the "mvano merged commit 71abc55" message is about. I filed a ticked about it: https://support.github.com/ticket/personal/0/1623119

PR #338, opened 2021-10-04
Sets FLAG_IMMUTABLE to the PendingIntent on FocusActivity

andreban:
Starting with API level 31, developers are required to set
FLAG_MUTABLE or FLAG_IMMUTABLE to PendingIntents. This pull
request makes android-browser-helper compatible to API level 31.

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/FocusActivity.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/FocusActivity.java
@@ -4,6 +4,7 @@ import android.app.Activity;
 import android.app.PendingIntent;
 import android.content.Context;
 import android.content.Intent;
+import android.os.Build;
 import android.os.Bundle;
 
 import androidx.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/FocusActivity.java
```diff
@@ -39,8 +40,12 @@ public static void addToIntent(Intent containerIntent, Context context) {
         // flag.
         focusIntent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
 
+        // Setting FLAG_MUTABLE or FLAG_IMMUTABLE is required from API Level 31 and above. However,
+        // the flag is not available below API level 23.
+        int flags =
```
PEConn, line 45: So in AndroidX we don't bother gating this behind a version check - we just use `PendingIntent.FLAG_IMMUTABLE` even though our `minSdkVersion` is `16` ([see here](https://source.corp.google.com/aosp-androidx/browser/browser/src/main/java/androidx/browser/customtabs/CustomTabsClient.java;l=232?q=FLAG_IMMUTABLE%20f:browser&sq=package:%5Eaosp-androidx$&type=cs)).

I'm not 100% sure how that all works (maybe since it's just a constant it just gets hardcoded in and then just has no effect on versions pre-M), but it does seem to work.

PEConn: I don't mind leaving it in (odd that the AndroidX lint doesn't complain though).

PEConn (approved):

PR #342, opened 2021-11-04
Updates dependencies to androidx.browser-1.4.0

andreban:
- Updates androidx.browser dependencies to the latest 1.4.0
- Updates demos androidbrowserhelper dependencies to 2.2.2
- Updates compileSdkVersion to 31, as it's required by androidx.browser-1.4.0

### DIFF
--- a/androidbrowserhelper/build.gradle
+++ b/androidbrowserhelper/build.gradle
@@ -51,7 +51,7 @@ android {
 }
 
 dependencies {
-    api 'androidx.browser:browser:1.3.0'
+    api 'androidx.browser:browser:1.4.0'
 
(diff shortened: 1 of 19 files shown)


PEConn:
Don't we also want to add android:exported in all of the manifests?

andreban:
Yes, but I'm doing another PR updating to targetSdk31 and adding exported (which is only required in that case).

PEConn (approved):
It would be nice to get all of the `buildToolsVersion`s aligned as well - although it's not necessary for this PR.

It would also be nice if we just defined these values once so we didn't have to update every file individually.

PR #349, opened 2022-01-05
Add version code to Digital Goods communications.

PEConn:
This is the first step to updating to support DGAPI v2.1.


PEConn:
We're OK developing on main right? Since these changes are only going into the playbilling target?

andreban (approved):
LGTM, thank you!

andreban:
> We're OK developing on main right? Since these changes are only going into the playbilling target?

Yep, developing in main is fine. If we need to release a bugfix, we can cherry pick into a tag

PR #350, opened 2022-01-05
Add type, iconUrl and introductoryPriceCycles to ItemDetails.

PEConn:
This is a fairly mechanical change, adding three new fields to ItemDetails.


andreban (approved):
LGTM, Thank you!

PR #352, opened 2022-01-06
Use queryPurchasesAsync instead of queryPurchases.

PEConn:
`queryPurchases` is deprecated, we should use `queryPurchasesAsync` instead.


EiraGe (approved):
LGTM

PR #353, opened 2022-01-06
Split some PurchaseDetails fields into LegacyPurchaseDetails.

PEConn:
We'll keep LegacyPurchaseDetails around to work with older versions of Chrome, but we want a the new PurchaseDetails class so that it can be shared by the ListPurchases call and the ListPurchaseHistoryCall.

### DIFF
--- a/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/LegacyPurchaseDetails.java
+++ b/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/LegacyPurchaseDetails.java
@@ -1,0 +1,112 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.androidbrowserhelper.playbilling.digitalgoods;
+
+import android.os.Bundle;
+
+import com.android.billingclient.api.Purchase;
+
+/**
+ * An old version of {@link PurchaseDetails} that was used for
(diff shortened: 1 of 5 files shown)


playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/LegacyPurchaseDetails.java
```diff
@@ -0,0 +1,112 @@
...
+    private static final String KEY_PURCHASE_STATE = "purchaseDetails.purchaseState";
+    private static final String KEY_PURCHASE_TIME_MICROSECONDS_PAST_UNIX_EPOCH =
+            "purchaseDetails.purchaseTimeMicrosecondsPastUnixEpoch";
+    private static final String KEY_WILL_AUTO_RENEW = "purchaseDetails.willAutoRenew";
+
+    static final int CHROMIUM_PURCHASE_STATE_UNKNOWN = 0;
+    static final int CHROMIUM_PURCHASE_STATE_PURCHASED = 1;
+    static final int CHROMIUM_PURCHASE_STATE_PENDING = 2;
+
+    /**
```
andreban, line 37: nit: this seems to be a left-over comment, since the ID is not stored here, but passed to the superclass on the constructor?

andreban (approved):
LGTM, with the one nit.

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/LegacyPurchaseDetails.java
```diff
@@ -0,0 +1,108 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
```
EiraGe, line 1: Should this be 2022?

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/LegacyPurchaseDetails.java
```diff
@@ -0,0 +1,108 @@
...
+
+package com.google.androidbrowserhelper.playbilling.digitalgoods;
+
+import android.os.Bundle;
+
+import com.android.billingclient.api.Purchase;
+
+/**
+ * An old version of {@link PurchaseDetails} that was used for
+ * {@link DigitalGoodsRequestHandler#DIGITAL_GOODS_API_VERSION} 1. This is kept around so that newer
```
EiraGe, line 23: This "1" seems unnecessary

EiraGe (approved):
LGTM with a couple nits

PR #400, opened 2022-11-25
Migrate notification sample

sebastianbenz:
The notification sample did not work as it didn't get migrated when raising the SDK target level.

### DIFF
--- a/demos/custom-tabs-example-app/src/main/java/org/chromium/customtabsdemos/NotificationParentActivity.java
+++ b/demos/custom-tabs-example-app/src/main/java/org/chromium/customtabsdemos/NotificationParentActivity.java
@@ -14,6 +14,7 @@
 
 package org.chromium.customtabsdemos;
 
+import android.app.NotificationChannel;
 import android.app.NotificationManager;
 import android.app.PendingIntent;
 import android.content.Context;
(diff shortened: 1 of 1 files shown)


demos/custom-tabs-example-app/src/main/java/org/chromium/customtabsdemos/NotificationParentActivity.java
```diff
@@ -69,8 +71,16 @@ public void onClick(View v) {
     }
 
     private void createAndShowNotification() {
+        NotificationManager notificationManager =
+                (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
+        NotificationChannel mChannel = notificationManager.getNotificationChannel(CT_NOTIFICATION_CHANNEL_ID);
```
PEConn, line 76: Don't we need to gate this behind a `Build.VERSION.SDK_INT` check, otherwise we'll crash on pre-O devices?

sebastianbenz: Min SDK is currently set to 26 which already supports NotificationChannel.

PEConn (approved):

demos/custom-tabs-example-app/src/main/java/org/chromium/customtabsdemos/NotificationParentActivity.java
sebastianbenz: I was surprised as well 🤷 I'll check with Android DevRel whether there's a recommended min sdk for samples.

PR #401, opened 2022-11-25
Fix Custom UI sample

sebastianbenz:
Actions were broken as PendingIntent has not been migrated after updated the target SDK.


andreban (approved):
LGTM

PR #405, opened 2023-02-06
Pull CustomTabsCallback to a getter from private field

mrajah1:
TWA based apps might want to listen to CCT call back events and take appropriate action ( eg. capture metrics about usage). Since the field currently is private and used in launchTWA() its not possible for the extending classes to change the behavior. This change pulls out the private field and creates a overridable method that supplies the Callback


PEConn (approved):
Change LGTM, thanks Mani.

mookey86 (approved):

PR #418, opened 2023-03-31
Use  WindowManager.LayoutParams.FLAG_FULLSCREEN to go full screen

mrajah1:
Argument to setFlags should be one of WindowManager.LayoutParams not a View Flag.


ifteneiftene (commented):
Ok

PR #423, opened 2023-05-06
Added column for share support in TWA

retog:
See https://developer.chrome.com/docs/android/trusted-web-activity/web-share-target/


PEConn:
Thanks for this change - I'll be happy to merge it, but you'll need to sign the CLA.

Issue #433, opened 2023-07-13
Bump the `com.google.android.gms:play-services-location` version in `locationdelegation`


andreban:
`com.google.android.gms:play-services-location`, used by the Location Delegation feature, includes a a transient dependency to `androidx.fragment` (via `play-services-base`), which is causing the Play Store to create a warning for applications using `android-browser-helper` + `locationdelegation` saying they are using an outdated SDK. See https://github.com/GoogleChromeLabs/bubblewrap/issues/810 for details.

Bumping the `play-services-location` in `locationdelegation` to the latest version should fix the issue.

SharmaDhruv2511:
I just have to change the location to change the issue?

PR #434, opened 2023-07-13
Updates dependencies for Location Delegation

andreban:
See #433


andreban:
Ugh, turns out core, locationdelegation and billing will need to be updated to `compileSdk 33`.

PR #435, opened 2023-07-18
Bump Billing module version

SayedElabady:
Bump Billing module version to 1.0.0-alpha10


PEConn (approved):
LGTM, noting that 1.0.0-alpha09 was published although the code was never updated.

PR #436, opened 2023-07-20
Fix background interactivity in partial custom tab

DordeDimitrijev:
Disabling background activity on custom partial tabs via checkbox wasn't working due the typo in the name of intent extra.


andreban (approved):
Thanks for the fix, @djoled97! LGTM

PR #444, opened 2023-10-03
Update a comment in PostMessage demo

SayedElabady:
SOURCE_ORIGIN gets checked to be a valid origin and that wasn't clear from the comment, right now the browserservices.Origin only supports HTTTP and HTTPS.


PEConn (approved):
Thanks!

PR #452, opened 2023-11-09
Update the flag when querying the Activities which can handle VIEW intent

rayworks:
Update the query flag to cover more installed browser apps (besides Chrome) which may support custom tabs.

### DIFF
--- a/androidbrowserhelper/build.gradle
+++ b/androidbrowserhelper/build.gradle
@@ -20,11 +20,11 @@ apply plugin: 'maven-publish'
 def VERSION = "2.5.0";
 
 android {
-    compileSdkVersion 31
+    compileSdkVersion 33
 
     defaultConfig {
         minSdkVersion 16
-        targetSdkVersion 31
+        targetSdkVersion 33
         versionCode 1
         versionName VERSION
 
--- a/build.gradle
+++ b/build.gradle
@@ -22,7 +22,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.1.1'
+        classpath 'com.android.tools.build:gradle:4.2.1'
     }
 }
 
--- a/demos/custom-tabs-example-app/build.gradle
+++ b/demos/custom-tabs-example-app/build.gradle
@@ -23,7 +23,7 @@ android {
     defaultConfig {
         applicationId "org.chromium.customtabsdemos"
         minSdkVersion 26
-        targetSdkVersion 31
+        targetSdkVersion 33
         versionCode 1
         versionName "1.0"
     }
--- a/demos/custom-tabs-example-app/src/main/java/org/chromium/customtabsdemos/CustomTabsHelper.java
+++ b/demos/custom-tabs-example-app/src/main/java/org/chromium/customtabsdemos/CustomTabsHelper.java
@@ -73,7 +73,7 @@ public class CustomTabsHelper {
         }
 
         // Get all apps that can handle VIEW intents.
-        List<ResolveInfo> resolvedActivityList = pm.queryIntentActivities(activityIntent, 0);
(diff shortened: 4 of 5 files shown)


PEConn:
Sorry, do you mind explaining a bit more what the change does?

rayworks:
Hi @PEConn , Because [only types of apps that are visible automatically](https://developer.android.com/training/package-visibility/automatic?hl=en#app-types-visible-automatically) since Android 11, by using the flag `PackageManager.MATCH_ALL`, we can get all candidates which are capable of handling intent with action `Intent.ACTION_VIEW`. After further check for the `CustomTabsService`, we won't miss other target browsers(e.g. org.mozilla.firefox) even if `Chrome` is not installed on the device.

PEConn:
The manifest for that demo app has:

```
    <queries>
        <intent>
            <action android:name="android.support.customtabs.action.CustomTabsService" />
        </intent>
    </queries>
```

So all browsers that support Custom Tabs should be visible.

rayworks:
Hi @PEConn ,
On my side, adding `<action android:name="android.support.customtabs.action.CustomTabsService" />` is not enough.
As I can tell from the testing result on OnePlus 7Pro (Android 12), the query result list only contains browser `Chrome`.
The remaining browsers get ignored somehow.

rayworks:
Query results changed after I updated the flag from `0` to `PackageManager.MATCH_ALL` at [this line of code](https://github.com/GoogleChrome/android-browser-helper/blob/main/demos/custom-tabs-example-app/src/main/java/org/chromium/customtabsdemos/CustomTabsHelper.java#L76).

BEFORE:
Screenshot 2023-11-13 at 11 27 52

AFTER:
Screenshot 2023-11-13 at 11 30 01

PEConn:
That's really weird - ultimately though I'm not against making this change, as it is in demo app code. Do you mind adding a comment saying why you're doing this and that in practice (name the device and Android version you're using) you've found the behaviour to be different from what's documented.

rayworks:
It turns out the method `queryIntentActivities(intent, flag)` has been deprecated on Android 13+, the new one is [queryIntentActivities (Intent intent, PackageManager.ResolveInfoFlags flags)](https://developer.android.com/reference/android/content/pm/PackageManager#queryIntentActivities(android.content.Intent,%20android.content.pm.PackageManager.ResolveInfoFlags)). 
As you can see from the values for [ResolveInfoFlags](https://developer.android.com/reference/android/content/pm/PackageManager.ResolveInfoFlags#of(long)), flag `0` is no long the option any more, so I'd rather use the `MATCH_ALL` in the code.

PEConn:
> As you can see from the values for [ResolveInfoFlags](https://developer.android.com/reference/android/content/pm/PackageManager.ResolveInfoFlags#of(long)), flag 0 is no longer the option any more, so I'd rather use the MATCH_ALL in the code.

Maybe I'm missing something, but the doc you linked says "long: Value is either 0 or a combination of ...".

Looked at your code, I'm fine with bumping the `compileSdkVersion` as long as you've run the tests. I'd still like you to add a the comment I mentioned [here](https://github.com/GoogleChrome/android-browser-helper/pull/452#issuecomment-1808176426).

rayworks:
@PEConn 
Please have another round of review.

PEConn:
Hey,

Thanks for the update. Sorry for being so picky, but could you please either:

- Undo the `compileSdkVersion` change (keeping it at 31).
- Or, update other usages of `queryIntentActivities` (such as in `ManageDataLauncherActivity` and `TwaProviderPicker`).

As it is, it's a little inconsistent that we've updated the code in one place we use `queryIntentActivities`, but not in others.

I'd prefer it if you took the first option and landed this PR with just a change to `CustomTabsHelper`, and then (if you're still interested) upload a separate PR that updates the `targetSdkVersion`. In general, small PRs that do single logical things are preferred.

Thanks,
Peter

Issue #461, opened 2024-03-04
Broken caching behavior in `SplashImageTransferTask` on dark / light mode switches


uwolfer:
In case you provide both normal (light) and dark (night) mode resources for your slash screen,` SplashImageTransferTask` does not work as expected: it keeps the splash screen for the previous mode. The only workaround is reinstalling the app.

**To Reproduce**
1. Add a folder res/drawable-night next to the existing res/drawable folder with a splash.png file.
2. Toggle the device color schema mode (light / dark).
3. The splash displayed shows still the "old" splash image.

I think I was able to find the cause of this issue:
https://github.com/GoogleChrome/android-browser-helper/blob/main/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/splashscreens/SplashImageTransferTask.java#L107 checks the last app update time, and compares it with the persisted time when the splash got last cached (`PREF_LAST_UPDATE_TIME`). That means for the case when you update the color schema of your device, but you do not reinstall the app, it will take the shortcut and use the already cached splash setup.

**Expected behavior**
Whenever the user changes color schema of their device, the proper splash should be rendered.

PR #463, opened 2024-03-10
Adapt caching behaviour for dark/light theme switching 

khmyznikov:
Trying to address [this bug](https://github.com/GoogleChrome/android-browser-helper/issues/461)

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/splashscreens/SplashImageTransferTask.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/splashscreens/SplashImageTransferTask.java
@@ -15,9 +15,11 @@
 package com.google.androidbrowserhelper.trusted.splashscreens;
 
 import android.annotation.SuppressLint;
+import android.app.UiModeManager;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.content.pm.PackageManager;
+import android.content.res.Configuration;
 import android.graphics.Bitmap;
 import android.os.AsyncTask;
 import android.util.Log;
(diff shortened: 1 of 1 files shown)


khmyznikov:
@uwolfer is it looks like a fix?

khmyznikov:
@PEConn @andreban how to check build?

andreban (changes requested):
Hey Gleb, thank you for the PR!

Would you mind adding how this change is solving #461 to the PR description? I also dropped a comment with regards moving magic values to static variables.

khmyznikov:
@andreban from my understanding, it solves the issue of wrong caching of different icons for light/dark splash screen. https://github.com/GoogleChrome/android-browser-helper/pull/51#issuecomment-1978621933

andreban:
Thanks for the update! What I'm making sure I understand is *how* this PR solves the issue - my understanding is that the bug is caused by only re-transferring the splash image file when the application is updated. But it also needs to be transferred when the theme changes. To solve that, you're adding the theme to the default preferences object, retrieving the value on launch, and ensuring the existing file is only re-used if all the following are true: the file exists, the app hasn't been updated, and the theme wasn't change. Does this sound right?

khmyznikov:
@andreban yeah that's sounds right 👍

andreban (approved):
LGTM - Will wait for @PEConn to take a look before merging.

uwolfer:
> @uwolfer is it looks like a fix?

Thanks for working on this, @khmyznikov. Code makes sense to me, but I have not tested it. I would suggest that you update the PR title to reflect what we actually have fixed with this change.

uwolfer:
@khmyznikov Do you think we can make progress on this PR?

khmyznikov:
@uwolfer sure, I've updated the branch 👍

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/splashscreens/SplashImageTransferTask.java
```diff
@@ -68,6 +72,25 @@ public SplashImageTransferTask(Context context, Bitmap bitmap, String authority,
...
+     * @return The current theme mode, either "light" or "dark".
+     */
+    private String getCurrentThemeMode(Context context) {
+        UiModeManager uiModeManager = (UiModeManager) context.getSystemService(Context.UI_MODE_SERVICE);
+        int nightModeFlags = context.getResources().getConfiguration().uiMode & Configuration.UI_MODE_NIGHT_MASK;
+        switch (nightModeFlags) {
+            case Configuration.UI_MODE_NIGHT_YES:
+                return "dark";
+            case Configuration.UI_MODE_NIGHT_NO:
+            case Configuration.UI_MODE_NIGHT_UNDEFINED:
```
uwolfer, line 88: Is it safe to handle those two equally?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/splashscreens/SplashImageTransferTask.java
```diff
@@ -68,6 +72,25 @@ public SplashImageTransferTask(Context context, Bitmap bitmap, String authority,
...
     }
 
+    /**
+     * Get the current theme mode.
+     *
+     * @param context Context to get the UiModeManager service.
+     * @return The current theme mode, either "light" or "dark".
+     */
+    private String getCurrentThemeMode(Context context) {
+        UiModeManager uiModeManager = (UiModeManager) context.getSystemService(Context.UI_MODE_SERVICE);
```
gstepniewski-google, line 82: This seems unused?

gstepniewski-google (changes requested):
Looks good, but needs a tiny bit of cleanup

PR #470, opened 2024-05-15
Fix source origin comment

SayedElabady:
The comment on the SOURCE_ORIGIN field in PostMessage demo is misleading, this PR fixes it.


demos/twa-post-message/src/main/java/com/google/androidbrowserhelper/demos/twapostmessage/MainActivity.java
```diff
@@ -56,8 +56,8 @@ public class MainActivity extends AppCompatActivity {
     private CustomTabsSession mSession;
     private Uri URL = Uri.parse("https://peconn.github.io/starters");
 
-    // This can be the app's package name, it has to either start with http or https.
-    private Uri SOURCE_ORIGIN = Uri.parse("https://my-app-origin-uri");
+    // This must be a valid origin, it has to either start with http or https.
```
PEConn, line 59: Probably worth emphasising here that the SOURCE_ORIGIN needs to have Digital Asset Links set up.

PEConn (approved):
Thanks!

PR #471, opened 2024-05-17
Move connecting/disconnecting from service to onCreate/Destroy

andreban:
**The problem**
I many cases, applications want to stay connected to the Custom Tabs service after the Custom Tabs is launched. They might want to listen for Custom Tabs callback events or enable a minimized tab to be replaced with a new one when relaunching. The existing code disconnects from the Custom Tabs service when the Activity is stopped, preventing either of those to happen.

**The solution**
Move binding and unbinding from the Custom Tabs Service to onCreate() / onDestroy(), so the application will stay connected to the service after it is backgrounded, allowing it to continue receiving callbacks and replace a minimized Custom Tab.

### DIFF
--- a/demos/custom-tabs-example-app/src/main/java/org/chromium/customtabsdemos/ServiceConnectionActivity.java
+++ b/demos/custom-tabs-example-app/src/main/java/org/chromium/customtabsdemos/ServiceConnectionActivity.java
@@ -15,6 +15,7 @@ package org.chromium.customtabsdemos;
 
 import android.net.Uri;
 import android.os.Bundle;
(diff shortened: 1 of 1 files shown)


PEConn (approved):
Code changes look good, but could you add a bit more to the description - you say what goes wrong, but not how this change fixes it.

andreban:
> Code changes look good, but could you add a bit more to the description - you say what goes wrong, but not how this change fixes it.

Updated the PR description. PTAL.

PEConn:
Amazing, LGTM, thanks.

PR #476, opened 2024-07-22
Update PlayBilling version to V6

orbital17:
WIP: API updates for Play Billing V6 migration

### DIFF
--- a/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/PaymentActivity.java
+++ b/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/PaymentActivity.java
@@ -63,6 +63,9 @@ public class PaymentActivity extends Activity implements BillingWrapper.Listener
             fail("Could not parse SKU.");
             return;
         }
+        if (mMethodData.isPriceChangeConfirmation) {
+            fail("Price change confirmation flow is not supported");
+        }
 
         /**
          * Note that we have temporarily disabled the IMMEDIATE_WITHOUT_PRORATION mode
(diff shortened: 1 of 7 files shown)


OlegNitz (approved):

OlegNitz (approved):

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/PaymentActivity.java
```diff
@@ -122,13 +122,7 @@ private void onSkusQueried(BillingResult result, List<SkuDetails> skus) {
...
 
     private void launchPriceChangeConfirmationFlow(SkuDetails sku) {
-        mWrapper.launchPriceChangeConfirmationFlow(this, sku, result -> {
-            if (result.getResponseCode() == BillingClient.BillingResponseCode.OK) {
-                setResultAndFinish(PaymentResult.priceChangeSuccess());
-            } else {
-                fail("Price change confirmation flow ended with result: " + result);
-            }
-        });
+        fail("Price change confirmation flow is not supported");
```
PEConn, line 125: I think we can fail earlier. At the moment, the `PaymentActivity` is launched, it connects to Play Billing (which is async) and queries the SKUs (which is async) then hits this, but we know in `PaymentActivity#onCreate` if the request is for a price change confirmation flow.

I think you could instead add `if (mMethodData.isPricechangeConfirmation) { fail(...) }` as soon as we create the MethodData.

PEConn (commented):
Looks good, just one thought.

PEConn (approved):

PEConn:
To check, you've run the tests locally and they still pass, right?

orbital17:
> To check, you've run the tests locally and they still pass, right?

Yes, tests pass locally

PR #486, opened 2024-12-03
CCT - AuthTab Sample Demo

erikrodriguez-se:
This PR contains a working demo for an implementation of AuthTab for Chrome Custom Tabs for Android. The included project, custom-tabs-authview, uses the existing custom-tabs-oauth demo as a guide for authentication using a website, in this case Github. 

Some notes: 
- The demo uses the custom redirect scheme. If it is desired to use https scheme instead, I can update the project accordingly. 
- Some updates beyond the demo were required for the android-browser-helper project to compile in newer versions of Android Studio. Namely, these changes included namespace additions to gradle files as well as updating the gradle version used in the project.


orbital17:
Hi Erik, thanks for the PR!
In general, SDK version, Java version and Gradle version need to be updated, but they need to be kept consistent across the project, and it's better to handle these updates in a separate PR. 
Would you like to create a PR for the updates? We are not sure when we’ll be able to handle this ourselves. Alternatively, you can try to implement the demo without updating the versions.

ukaratkevich:
+1 to make a separate PR with build changes. 
I also doubt that we want to use androidx snapshots for demo, I'd wait at least for 1.9.0 alpha.

erikrodriguez-se:
Hi all, I can split this in to two PRs. I'll make this one specific to AuthTab and will create another for project updates.

demos/custom-tabs-authview/src/main/java/com/google/androidbrowserhelper/demos/customtabsauthview/AuthManager.java
```diff
@@ -0,0 +1,99 @@
...
+        Uri uri = Uri.parse(mAuthorizationEndpoint)
+                .buildUpon()
+                .appendQueryParameter("response_type", "code")
+                .appendQueryParameter("client_id", mClientId)
+                .appendQueryParameter("scope", scope)
+                .appendQueryParameter("state", state)
+                .build();
+
+        // Open the Authorization URI in a Chrome Custom Auth Tab.
+        AuthTabIntent authTabIntent = new AuthTabIntent.Builder().build();
```
ukaratkevich, line 62: I see that AuthView demo is almost identical to the Oauth demo, the difference are:
1. New callback intent-filter
2. Launcher intents are different
3. Different result handling

I think that we can merge this demo into Oauth, as far as I can see it'd require:
1.  Add an AuthView callback intent-filter into the AndroidManifest
2. Demo should provide a way to switch between CCT and AuthView. I think a checkbox and `shouldUseAuthView` should be enough.
3. In AuthManager launch intents and handle results based on `shouldUseAuthView`.
4. I'd also add comments why AuthView is preferable in AuthManager or even deprecate CCT flow at all.

This way we get rid of duplication and provide some kind of migration steps from CCT to AuthView.
What do you think?

erikrodriguez-se: Hi @ukaratkevich, 

You're right that these two demos are very similar. The new Auth Tab implementation is intended to be a replacement for the older Oauth implementation strategy. The Oauth project uses standard CCT. The new project uses Auth Tab but also leverages the older CCT method as a fallback. So there are a lot of pieces between the two demos that are identical. I was asked not to delete the Oauth project but instead to keep it for historical purposes. 

As far as what to do next, it's important to know that Auth Tab *is* CCT, and so we shouldn't really be representing a choice between one or the other to the developer. Instead, the demo shows the new Auth Tab implementation with the standard CCT as a graceful fallback, which is I think is the flow we want to demonstrate to developers. Adding booleans and checkboxes defeats that intention. 

I think perhaps it would be easiest to delete the older Oauth demo entirely, if that is creating confusion. Would that help?

ukaratkevich: Thanks a lot for the explanation, in this case I'd not delete the oath demo, but rather mark it as deprecated in the MD file with `@Deprecated` annotation on main classes and add a comment to use Auth View instead.

Overall, lgtm :)

ukaratkevich:
Hi @erikrodriguez-se, please, resolve all merge conflicts and we will begin to approve all changes. Thank!

ukaratkevich (approved):

ukaratkevich:
Hi @erikrodriguez-se, there's a new merge conflict, also, please check local build errors, because R class is unresolved and AndroidManifest has errors.

demos/custom-tabs-auth-tab/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,35 @@
...
+                <action android:name="android.intent.action.MAIN" />
+                <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
+
+            <!-- An intent-filter for fallback to Chrome Custom Tabs -->
+            <intent-filter android:autoVerify="true">
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="auth" />
```
ukaratkevich, line 30: Lint requires to have http or https according to https://developer.android.com/training/app-links#android-app-links.
Does custom scheme work?

demos/custom-tabs-auth-tab/src/main/java/com/google/androidbrowserhelper/demos/customtabsauthview/MainActivity.java
```diff
@@ -0,0 +1,101 @@
...
+    private TextView mUserText;
+    private ProgressBar mProgressBar;
+    private boolean mLoggedIn;
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_main);
+
+        mLoginButton = findViewById(R.id.login_button);
```
ukaratkevich, line 58: R class import is missing

ukaratkevich (changes requested):

demos/custom-tabs-auth-tab/src/main/AndroidManifest.xml
erikrodriguez-se: @ukaratkevich This is corrected, it should not have had the "autoVerify" attribute.

ukaratkevich: Perfect, thanks a lot! Also big appreciation for removing snapshot repos.

PR #497, opened 2025-03-06
DO NOT MERGE YET Protocol Handler support

gstepniewski-google:
Waiting for a new version of androidx.browser to become available, so that we can use the new TrustedWebActivityIntentBuilder method in LauncherActivity#221.


demos/twa-firebase-analytics/src/main/java/com/google/androidbrowserhelper/demos/twa_firebase_analytics/FirebaseAnalyticsLauncherActivity.java
```diff
@@ -60,12 +58,14 @@ protected boolean shouldLaunchImmediately() {
     }
 
     @Override
-    protected Uri getLaunchingUrl() {
-        Uri uri = super.getLaunchingUrl();
+    protected LaunchUrlBundle getLaunchUrlBundle() {
```
orbital17, line 61: @gstepniewski-google I would rather not change this method. Can we  leave it like it was and just check in launchTwa if protocol handlers are applied?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -339,26 +347,54 @@ public void onEnterAnimationComplete() {
...
-        Uri uri = getIntent().getData();
-        if (uri != null) {
-            Log.d(TAG, "Using URL from Intent (" + uri + ").");
-            return uri;
-        }
-
-        if (mMetadata.defaultUrl != null) {
-            Log.d(TAG, "Using URL from Manifest (" + mMetadata.defaultUrl + ").");
-            return Uri.parse(mMetadata.defaultUrl);
+    protected LaunchUrlBundle getLaunchUrlBundle() {
```
ukaratkevich, line 376: We're breaking compatibility here, I'd suggest to keep `getLaunchingUrl` and deprecate it if we prefer new API. If we don't do that, then we must release 3.0 ABH which I'd prefer to avoid.

demos/twa-firebase-analytics/src/main/java/com/google/androidbrowserhelper/demos/twa_firebase_analytics/FirebaseAnalyticsLauncherActivity.java
ukaratkevich: Generally, I agree with @orbital17 that we should not change it.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -330,6 +341,23 @@ protected void onSaveInstanceState(Bundle outState) {
...
+    /**
+     * Override this to enable Protocol Handler support.
+     * Keys of this map are data schemes, e.g. "bitcoin", "irc", "xmpp", "web+coffee", and values
+     * are templates that will be used to construct the full URL. The template must contain a "%s"
+     * token and be an absolute location with http/https scheme and the same origin as the TWA.
+     *
+     * An example valid entry in the map would be:
+     * ["web+coffee"] -> ["https://coffee.com/?type=%s"]
+     * This would result in a link "web+coffee://latte" being converted to
+     * "https://coffee.com/?type=latte".
```
glennhartmann, line 353: My understanding of the spec is that the protocol should be included in the replacement of "%s". Ie, I think the link should be converted to "https://coffee.com/?type=web%2Bcoffee%3A%2F%2Flatte"

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -347,18 +375,27 @@ public void onEnterAnimationComplete() {
...
+        Uri baseUrl = Uri.parse(mMetadata.defaultUrl);
+
+        Uri intentUrl = getIntent().getData();
+
+        if (intentUrl != null) {
+            Map<String, Uri> protocolHandlers = getProtocolHandlers();
+            String scheme = intentUrl.getScheme();
+            if (protocolHandlers.containsKey(scheme)) {
+                String format = Objects.requireNonNull(protocolHandlers.get(scheme)).toString();
+                String target = intentUrl.toString().replace(scheme + "://", "");
```
glennhartmann, line 387: In accordance to my above comment, I don't _think_ you want to be stripping the scheme here.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -347,18 +375,27 @@ public void onEnterAnimationComplete() {
...
+                Uri targetUrl =  Uri.parse(String.format(format, target));
```
glennhartmann, line 388: You may need to percent-encode `target` before inserting it into `format`

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -347,18 +375,27 @@ public void onEnterAnimationComplete() {
...
-            return Uri.parse(mMetadata.defaultUrl);
+        Uri baseUrl = Uri.parse(mMetadata.defaultUrl);
+
+        Uri intentUrl = getIntent().getData();
+
+        if (intentUrl != null) {
+            Map<String, Uri> protocolHandlers = getProtocolHandlers();
+            String scheme = intentUrl.getScheme();
+            if (protocolHandlers.containsKey(scheme)) {
+                String format = Objects.requireNonNull(protocolHandlers.get(scheme)).toString();
```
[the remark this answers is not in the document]
gstepniewski-google: I don't think it would have, because we're checking if the key exists in the map one line before and the `requireNonNull` is only there to keep lint happy... (unless someone is doing something funny with the map they're returning to us like modifying it from another thread)

Still, printing a warning is a good idea so I've added it and did a small refactor to get rid of the potential exception.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
gstepniewski-google: You're right! Not sure how we all missed it but thank you for noticing!

PR #500, opened 2025-04-14
Bumped all dependencies necessary to use newest androidx.browser 1.9.0-alpha02

gstepniewski-google:


ukaratkevich (approved):

Aaron-Ritter:
Hi @gstepniewski-google,

We had a failing dependency update due to the minimum API lvl 36 introduced by this PR. And I’m trying to understand the reason to bump to an alpha version and minimum API lvl 36 dependency and packing it in to a release?

Since the newest Android version is not released, should this at least be added in to a pre-release of [android-browser-helper](https://github.com/GoogleChrome/android-browser-helper) instead?

PR #501, opened 2025-04-14
Protocol Handler support

gstepniewski-google:


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
```diff
@@ -347,18 +374,33 @@ public void onEnterAnimationComplete() {
...
+
+            Uri format = protocolHandlers.get(scheme);
+            if (format != null) {
+                String target = Uri.encode(intentUrl.toString());
+                Uri targetUrl =  Uri.parse(String.format(format.toString(), target));
+                Log.d(TAG, "Using protocol handler url: " + targetUrl);
+                return targetUrl;
+            }
+
+            if ("https".equals(scheme)) {
```
ukaratkevich, line 393: nit: I'd move it before `format != null`, because this is the default case that we expect.

ukaratkevich (approved):

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
gstepniewski-google: Good point!

PR #513, opened 2025-05-09
Use a task id as a session id

SelyvonSerhii:
Since we are going to have several tasks of the same TWA in parallel we'll need to have separate sessions for each task. Therefore we will use task id value which is a hex number for session  id instead.

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
@@ -34,11 +34,12 @@ import androidx.browser.trusted.Token;
 import androidx.browser.trusted.TokenStore;
 import androidx.browser.trusted.TrustedWebActivityIntent;
 import androidx.browser.trusted.TrustedWebActivityIntentBuilder;
-import androidx.core.content.ContextCompat;
 
 import com.google.androidbrowserhelper.BuildConfig;
-import com.google.androidbrowserhelper.trusted.ChromeOsSupport;
-import com.google.androidbrowserhelper.trusted.splashscreens.SplashScreenStrategy;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Random;
 
 /**
  * Encapsulates the steps necessary to launch a Trusted Web Activity, such as establishing a
@@ -47,7 +48,7 @@ import com.google.androidbrowserhelper.trusted.splashscreens.SplashScreenStrateg
 public class TwaLauncher {
     private static final String TAG = "TwaLauncher";
 
-    private static final int DEFAULT_SESSION_ID = 96375;
+    private static final Map<Integer, Integer> mTaskIdToSessionId = new HashMap<>();
 
     private static final String EXTRA_STARTUP_UPTIME_MILLIS =
             "org.chromium.chrome.browser.customtabs.trusted.STARTUP_UPTIME_MILLIS";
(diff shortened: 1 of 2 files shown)


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -107,17 +108,17 @@ void launch(Context context,
      * Creates an instance that will automatically choose the browser to launch a TWA in.
      * If no browser supports TWA, will launch a usual Custom Tab (see {@link TwaProviderPicker}.
      */
-    public TwaLauncher(Context context) {
-        this(context, null);
+    public TwaLauncher(Activity activity) {
```
ukaratkevich, line 111: Changing constructor parameters breaks API, because `Context` is broader than `Activity`, e.g. consumers might pass `ContextWrapper` or `getApplicationContext` here.  Some concerns:

1. Provide a new constructor for narrowed `Activity` scope. If passing in `Activity` is mandatory, then we need to deprecate existing constructors.
2. Can we backport constructor with `Context` somehow? If we deprecate it, it might mess up sessions ids.
3. Can we not use task ids? I have security concerns here, because this information should not be trusted to 3p app. Can we have a static counter for sessions instead?

ukaratkevich: @gstepniewski-google  for thoughts too.

SelyvonSerhii: It's not necessary to pass an Activity. I can add one more constructor with additinal param and pass just int value for taskId.  
Regarding TaskId usage itself. Currently we use hardcodded value, it's also not secure. What kind of new security concern do you mean about changing it to taskId?

gstepniewski-google: I agree that we cannot change Context to Activity, there are many many things that are a Context but not an Activity. 
I'm not entirely sure what is happening here but it looks like `DEFAULT_SESSION_ID` was just a random number so we should probably just keep using random numbers?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -116,27 +117,27 @@ void launch(Context context,
      * Creates an instance that will automatically choose the browser to launch a TWA in.
      * If no browser supports TWA, will launch a usual Custom Tab (see {@link TwaProviderPicker}.
      */
-    public TwaLauncher(Context context) {
```
gstepniewski-google, line 119: I think we should leave this constructor available, otherwise we might be breaking compatibility?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -149,6 +150,19 @@ public TwaLauncher(Context context, @Nullable String providerPackage, int sessio
         }
     }
 
+    static private Integer makeSessionId(int taskId) {
```
gstepniewski-google, line 153: nit: non-canonical modifier order (should be `private static`)

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
SelyvonSerhii: Makes sense

SelyvonSerhii: Done

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -120,12 +121,20 @@ public TwaLauncher(Context context) {
...
+        this(context, taskId, null);
+    }
+
     /**
      * Same as above, but also allows to specify a browser to launch. If specified, it is assumed to
      * support TWAs.
      */
-    public TwaLauncher(Context context, @Nullable String providerPackage) {
-        this(context, providerPackage, DEFAULT_SESSION_ID,
+    public TwaLauncher(Context context, Integer taskId, @Nullable String providerPackage) {
```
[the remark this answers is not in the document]
SelyvonSerhii: Done

gstepniewski-google (approved):

PR #515, opened 2025-05-13
Add metadata option to control Chrome startup timing relative to anim…

youenn98:
Currently, the biggest bottleneck of app start up time is that the TWA app won't start the chrome until the enter animation is complete. For some Apps, it could take 500ms to play the animation and chrome just do nothing. It could help improve user experience if we can configure to start chrome early before enter animation complete.

This commit adds a new metadata option,
`START_CHROME_BEFORE_ANIMATION_COMPLETE`, to LauncherActivityMetadata.

This option allows developers to control whether Chrome starts before or after the TWA enter animation completes. The current and default behavior is to start Chrome after the enter animation finishes, ensuring a smooth transition.


PEConn (approved):
LGTM, thanks! @OlegNitz for approvals as well.

PR #517, opened 2025-05-15
Add an extra data for TWA startup timestamp

joehattori:
Add an extra data for the timestamp for TWA startup. This extra data will be passed to the browser to calculate various metrics e.g. the duration from TWA launch to FCP.

### DIFF
--- a/androidbrowserhelper/build.gradle
+++ b/androidbrowserhelper/build.gradle
@@ -29,6 +29,8 @@ android {
         versionCode 1
         versionName VERSION
 
+        buildConfigField "int", "LIBRARY_VERSION", "${versionCode}"
+
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
     }
 
@@ -37,6 +39,10 @@ android {
             minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
+    }
+
+    buildFeatures {
+        buildConfig true
     }
 
     compileOptions {
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
@@ -36,6 +36,7 @@ import androidx.browser.trusted.TrustedWebActivityIntent;
 import androidx.browser.trusted.TrustedWebActivityIntentBuilder;
 import androidx.core.content.ContextCompat;
 
+import com.google.androidbrowserhelper.BuildConfig;
 import com.google.androidbrowserhelper.trusted.ChromeOsSupport;
 import com.google.androidbrowserhelper.trusted.splashscreens.SplashScreenStrategy;
 
@@ -47,6 +48,12 @@ public class TwaLauncher {
     private static final String TAG = "TwaLauncher";
 
     private static final int DEFAULT_SESSION_ID = 96375;
+
+    private static final String EXTRA_STARTUP_UPTIME_MILLIS =
+            "org.chromium.chrome.browser.customtabs.trusted.STARTUP_UPTIME_MILLIS";
+
+    private static final String EXTRA_ANDROID_BROWSER_HELPER_VERSION =
+            "org.chromium.chrome.browser.ANDROID_BROWSER_HELPER_VERSION";
 
     public static final FallbackStrategy CCT_FALLBACK_STRATEGY =
             (context, twaBuilder, providerPackage, completionCallback) -> {
@@ -95,6 +102,8 @@ public class TwaLauncher {
     private TokenStore mTokenStore;
 
     private boolean mDestroyed;
+
+    private long mStartupUptimeMillis;
 
     public interface FallbackStrategy {
         void launch(Context context,
@@ -269,6 +278,11 @@ public class TwaLauncher {
         }
         Log.d(TAG, "Launching Trusted Web Activity.");
         TrustedWebActivityIntent intent = builder.build(mSession);
+        if (mStartupUptimeMillis != 0) {
+            intent.getIntent().putExtra(EXTRA_STARTUP_UPTIME_MILLIS, mStartupUptimeMillis);
+        }
+        intent.getIntent().putExtra(
+                EXTRA_ANDROID_BROWSER_HELPER_VERSION, BuildConfig.LIBRARY_VERSION);
         FocusActivity.addToIntent(intent.getIntent(), mContext);
         intent.launchTrustedWebActivity(mContext);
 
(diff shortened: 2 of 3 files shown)


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -48,6 +49,9 @@ public class TwaLauncher {
 
     private static final int DEFAULT_SESSION_ID = 96375;
 
+    private static final String EXTRA_STARTUP_UPTIME_MILLIS =
+            "android.support.customtabs.trusted.STARTUP_UPTIME_MILLIS";
```
PEConn, line 53: nit: I'd probably not have this be `android.support.customtabs` as it's not in (and not going to be in) AndroidX (which used to be the Android Support Library). Since it's only going to be read by Chromium, `org.chromium.chrome.browser.customtabs.trusted.STARTUP_TIME_MILLIS` or something should be good.

joehattori: Thank you for the review. Applied to the new commit.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -269,6 +273,7 @@ private void launchWhenSplashScreenReady(TrustedWebActivityIntentBuilder builder
         }
         Log.d(TAG, "Launching Trusted Web Activity.");
         TrustedWebActivityIntent intent = builder.build(mSession);
+        intent.getIntent().putExtra(EXTRA_STARTUP_UPTIME_MILLIS, SystemClock.uptimeMillis());
```
ishitatsuyuki, line 276: I think we should start measuring the timestamp when the activity starts, not when we launch the intent (this would otherwise fail to measure the animation delay); so we should try to pass a timestamp from the Activity. taking a timestamp in launch() might be also fine if it's close enough.

joehattori:
@PEConn Thanks for the review. I included the code to pass the version of ABH to the browser in the new version. If you think this should be another PR, please let me know.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -107,28 +115,29 @@ void launch(Context context,
      * Creates an instance that will automatically choose the browser to launch a TWA in.
      * If no browser supports TWA, will launch a usual Custom Tab (see {@link TwaProviderPicker}.
      */
-    public TwaLauncher(Context context) {
-        this(context, null);
+    public TwaLauncher(Context context, long startupUptimeMillis) {
```
ishitatsuyuki, line 118: I think these are public APIs, and adding a new required parameter breaks API compatibility.

I think we might want to grab the timestamp in launch() instead for now, should be close enough.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -48,6 +49,11 @@ public class TwaLauncher {
 
     private static final int DEFAULT_SESSION_ID = 96375;
 
+    private static final String EXTRA_STARTUP_UPTIME_MILLIS =
+            "org.chromium.chrome.browser.customtabs.trusted.STARTUP_UPTIME_MILLIS";
+
+    private static final String EXTRA_ABH_VERSION = "org.chromium.chrome.browser.ABH_VERSION";
+
```
ishitatsuyuki, line 56: As in https://chromium-review.googlesource.com/c/chromium/src/+/6520581/comment/5794ab53_0454b740/, it might be better to use lower case naming conventions for the Intent strings.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
PEConn: Good catch about the public API! Normally what we'd do in this case is add a new overload that takes the value and then make the old version call that overload with a default value.

However, at the moment, both the constructor and `launch` have quite a few overloads already so things would get quite messy.

It's probably simplest to just have a `setStartupTimeMillis` method on `TwaLauncher`.

joehattori: Thanks for the review, I added `setStartupTimeMillis` in `TwaLauncher`.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
joehattori: As the corresponding variable in the CL moved to a different file, decided to leave the letters capitalized following its convention.

PEConn (approved):
LGTM, thanks

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -299,6 +313,13 @@ public String getProviderPackage() {
         return mProviderPackage;
     }
 
+    /**
+     * Sets the startup timestamp.
+     */
+    public void setTwaStartupUptimeMillis(long startupUptimeMillis) {
```
ishitatsuyuki, line 319: nit: I think `Twa` can be omitted since the class is already TWA-specific (TwaLauncher)

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -299,6 +313,13 @@ public String getProviderPackage() {
         return mProviderPackage;
     }
 
+    /**
+     * Sets the startup timestamp.
```
ishitatsuyuki, line 317:
```suggestion
     * Sets the timestamp (in SystemClock.uptimeMillis()) when the TWA launcher activity was created. This timestamp is used to report the full startup duration to the browser.
```

androidbrowserhelper/build.gradle
```diff
@@ -29,6 +29,8 @@ android {
         versionCode 1
```
ishitatsuyuki, line 29: non-blocking: we should remember to bump the versionCode for the next release.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -48,6 +49,12 @@ public class TwaLauncher {
 
     private static final int DEFAULT_SESSION_ID = 96375;
 
+    private static final String EXTRA_STARTUP_UPTIME_MILLIS =
+            "org.chromium.chrome.browser.customtabs.trusted.STARTUP_UPTIME_MILLIS";
+
+    private static final String EXTRA_ANDROID_BROWSER_HELPER_VERSION =
+            "org.chromium.chrome.browser.ANDROID_BROWSER_HELPER_VERSION ";
```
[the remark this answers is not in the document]
joehattori: Didn't notice that, thanks!

OlegNitz (approved):

PR #519, opened 2025-05-21
Fix the androidTest build

youenn98:
1. Remove package namespace definition in Manifest. Setting the namespace via the package attribute in the source AndroidManifest.xml is no longer supported.

2. Specify an explicit value for `android:exported` when the component has an intent filter defined.

3. Add androidx.appcompat:appcompat build dependency in build.gradle

4. Complete the PwaWrapperSplashScreenStrategy()'s startChromeBeforeAnimationComplete argument in test


ukaratkevich (approved):
Thanks for the update!

PR #520, opened 2025-05-22
Skip animation as default

youenn98:
Make start chrome before animation complete as default behavior. Helps improve twa start-up time.

### DIFF
--- a/androidbrowserhelper/build.gradle
+++ b/androidbrowserhelper/build.gradle
@@ -17,7 +17,11 @@
 apply plugin: 'com.android.library'
 apply plugin: 'maven-publish'
 
-def VERSION = "2.6.0";
+// Before 2.6.1, the version code was unused and was kept at 1.
+// This is now being used to report metrics, and should be bumped with
+// every version bump.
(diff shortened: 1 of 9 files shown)


ishitatsuyuki (changes requested):
From offline discussion, it seems PwaWrapperSplashScreenStrategyTest.java is still testing with startChromeBeforeAnimationComplete = false? I think we should ensure the defaults are tested, and if waitsForEnterAnimationCompletion_BeforeDeclaringReady breaks, we should state in the changelog that the API contract is changing.

youenn98:
Yes, if we change the behavior to not wait on animation complete, the test waitsForEnterAnimationCompletion_BeforeDeclaringReady is not so meaningful. Added commits to delete it from current test suite.

OlegNitz (approved):

maiconcarraro:
@youenn98 do you know why it has a dedicated configuration for Chrome, but no preference for the TWA Picker to prefer Chrome? A lot of Samsung devices have Samsung internet by default w/ multiple issues with start time (esp. One UI 5.1)

PR #525, opened 2025-05-23
fix(web-view-fallback-activity): limit deprecated apis to android 14

dargmuesli:
Not sure if this is the right approach, but this is my guess to resolve on error reported in https://github.com/pwa-builder/PWABuilder/issues/4863, namely:

image


rafaeldev (commented):
Hi!

I don't have expertise in Android environment, but I have a doubt! When the version is greater than UPSIDE_DOWN_CAKE, is it not necessary to call `setStatusBarColor`? In this case, we don't need call other method?

Thanks

dargmuesli:
@rafaeldev thank you for reviewing! I think with edge-to-edge display there is no such thing as a status bar color to be set, but I might be wrong.

OlegNitz (approved):

PR #384, opened 2022-09-25
Merge recent changes into stable

EiraGe:
It's been a while since the stable branch gets updated. Merge all new changes release 2.4.0 to `stable` so they will be copy into the internal repo

PR #527, opened 2025-05-28
Update `stable` branch up to release 2.6.1

glennhartmann:
This will trigger copybaray to copy the changes into the internal repo (see https://github.com/GoogleChrome/android-browser-helper/pull/384)

### DIFF
--- a/androidbrowserhelper/build.gradle
+++ b/androidbrowserhelper/build.gradle
@@ -17,7 +17,11 @@
 apply plugin: 'com.android.library'
 apply plugin: 'maven-publish'
 
+// Before 2.6.1, the version code was unused and was kept at 1.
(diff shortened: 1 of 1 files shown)


glennhartmann:
The CLA check failing is a little bit odd, but I think we're probably safe to ignore it since we're just merging from `main` to `stable`, not actually adding any new changes.

I see one of the "missing CLA" users, for example, is @EiraGe. At the time of the change, she was covered under the CLA by virtue of being a Google employee. I can't personally vouch for the other (@erikrodriguez-se), but I imagine it's a similar story.

glennhartmann:
I've filed http://b/420907817 internally to double-check whether it's ok to ignore the CLA failure

PR #529, opened 2025-06-05
Add package name constructor to TwaLauncher

SelyvonSerhii:


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -121,31 +117,23 @@ public TwaLauncher(Context context) {
...
+        this(context, providerPackage, DEFAULT_SESSION_ID,
                 new SharedPreferencesTokenStore(context));
     }
 
     /**
      * Same as above, but also accepts a session id. This allows to launch multiple TWAs in the same
      * task.
      */
-    public TwaLauncher(Context context, @Nullable String providerPackage, @Nullable Integer taskId,
+    public TwaLauncher(Context context, @Nullable String providerPackage, int sessionId,
```
ukaratkevich, line 133: Still breaks compatibility, `int` is sticker than `@Nullable Integer`. I'd keep the new version since someone has already started to use the library.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -121,31 +117,23 @@ public TwaLauncher(Context context) {
         this(context, null);
     }
 
-    /**
-     * Same as above, but also allows to specify a task id to distinguish several sessions running
-     * for the same TWA app.
-     */
-    public TwaLauncher(Context context, @Nullable Integer taskId) {
```
ukaratkevich, line 128: We can't remove methods, because someone has been using it, I think we can keep it as is with my `@Nullable` proposal below. Another approach is to deprecate it and remove in the next release.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
SelyvonSerhii: Changed to Integer

ukaratkevich (approved):

PR #530, opened 2025-06-06
Use Guava's ImmutableMap.of() instead of Map.of().

glennhartmann:


dancerj (commented):
Nice, thank you.

ukaratkevich (approved):
Thanks!

PR #536, opened 2025-06-23
BillingClient library updated to 7.1.1

gstepniewski-google:


OlegNitz (approved):

playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/provider/MethodData.java
```diff
@@ -102,9 +104,16 @@ private static MethodData fromJsonInner(String json) throws JSONException {
...
+            replacementMode = getProration(dataObject);
+            if (replacementMode != null) {
+                Log.w("TAG", "prorationMode is deprecated, please migrate to replacementMode!");
+            }
+        }
+
+        return new MethodData(sku, isPriceChangeConfirmation, oldSku, purchaseToken, replacementMode);
     }
 
     @Nullable
```
rsolomakhin, line 119: Minor nit / drive-by:

It may be helpful to add a comment in `MethodData.fromIntent()` method that it is making an assumption that`method` (from `String method = methods.get(0)`) equals to `"https://play.google.com/billing"`.

PR #539, opened 2025-06-30
Use unique name for the string resource "app_name".

TorstenSommerfeld:
Having "app_name" is quite common and it can cause conflicts in apps that want to use this library or other libraries those app depend on.

This change will avoid conflicts with other libraries or apps.


androidbrowserhelper/src/main/res/values/strings.xml
```diff
@@ -1,5 +1,5 @@
 <resources>
-    <string name="app_name">Android Browser Helper</string>
+    <string name="android_browser_helperi_app_name">Android Browser Helper</string>
```
gstepniewski-google, line 2: nit:

```suggestion
    <string name="android_browser_helper_app_name">Android Browser Helper</string>
```

PR #551, opened 2025-09-18
Bump version to 2.7.0-alpha01

adamraine:
This creates a new alpha version, GH tag/release to be done once this lands.

An alpha release is necessary to publish changes dependent on the alpha release of `androidx.browser`.

Publishing to maven will need to be done by a repo owner

cc @gstepniewski-google

### DIFF
--- a/androidbrowserhelper/build.gradle
+++ b/androidbrowserhelper/build.gradle
@@ -22,7 +22,7 @@ plugins {
 // Before 2.6.1, the version code was unused and was kept at 1.
 // This is now being used to report metrics, and should be bumped with
 // every version bump.
-def VERSION = "2.6.2";
+def VERSION = "2.7.0-alpha01";
 def VERSION_CODE = 2;
 
 android {


androidbrowserhelper/build.gradle
```diff
@@ -22,7 +22,7 @@ plugins {
 // Before 2.6.1, the version code was unused and was kept at 1.
 // This is now being used to report metrics, and should be bumped with
 // every version bump.
-def VERSION = "2.6.2";
+def VERSION = "2.7.0-alpha01";
 def VERSION_CODE = 2;
```
adamraine, line 26: Should `VERSION_CODE` be bumped for alpha releases?

adamraine: > 2070001

Not sure what the format is here. Is the `0` between `2` and `7` just a delimiter?

For now, I would prefer to just bump to 3 if there isn't a pressing need to establish a format.

gstepniewski-google: My reasoning was two digits per version number, i.e. "02.07.00.01" but without the dots and the leading zero... but I'm probably overengineering at this point, let's just bump it to 3 as you're saying.

PR #562, opened 2025-11-16
feat(twa-launcher): allow adding extra to twa intent

### LINKED ISSUE #329: Custom headers with TWALauncher
Hi,

I have built a TWA app that uses the [TwaLauncher](https://github.com/GoogleChrome/android-browser-helper/blob/main/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java) to launch.

I'm currently in a position where I need to send a token that was stored in a previous version of the native app to the web site which I'm starting. I thought the most suitable way to do this was to send it as a custom header.

So I took a look at the [Custom Headers Demo](https://github.com/GoogleChrome/android-browser-helper/tree/main/demos/custom-tabs-headers) and eventually got it working in my app.

However, the only way forward I could see was to copy the code from TwaLauncher into my own app and modify it.
Besides adding the `handle_all_urls` relationship to my asset statements, it seems as if `client.warmup` is always required as well as `mSession.validateRelationship(CustomTabsService.RELATION_USE_AS_ORIGIN, URL, null)`

To avoid having to "own" all the TwaLauncher code, I guess I'm requesting some way to tell it to set up the extra validation for me. Or am I missing something?

Best Regards
Tobias

dargmuesli:
Allows setting custom headers for TwaLauncher.
Resolves https://github.com/GoogleChrome/android-browser-helper/issues/329

See implementation example in:
- https://github.com/maevsi/android/pull/44

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
@@ -355,7 +355,7 @@ public class TwaLauncher {
                      // for further details.
         }
         Log.d(TAG, "Launching Trusted Web Activity.");
-        TrustedWebActivityIntent intent = builder.build(mSession);
+        TrustedWebActivityIntent intent = onPrepareIntent(builder.build(mSession));
         if (mStartupUptimeMillis != 0) {
             intent.getIntent().putExtra(EXTRA_STARTUP_UPTIME_MILLIS, mStartupUptimeMillis);
         }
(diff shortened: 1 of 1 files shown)


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -366,6 +366,10 @@ private void launchWhenSplashScreenReady(TrustedWebActivityIntentBuilder builder
         }
     }
 
+    protected TrustedWebActivityIntent getIntent(TrustedWebActivityIntent intent) {
```
gstepniewski-google: Please add javadoc explaining what this method does and how it should be used.
Also consider changing the name to something a little more fitting: `getIntent` is a method defined by many Android components and used to obtain the starting Intent - meanwhile this one is meant to allow subclasses to *customize* the Intent before passing it back.

gstepniewski-google (changes requested):

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
dargmuesli: Done :raised_hands: let me know what you think! (please feel free to change the name on your discretion if you have a better one in mind)

gstepniewski-google (approved):

PR #575, opened 2026-03-26
Add a dialogue for when custom tab fallback fails.

ericsum26:
This applies when there is no package available for any of the options (browser, trusted web activity, or custom tab). When the custom tab fallback itself fails because there is no provider available, there is currently an exception thrown somewhere in the Android package management code. Visually, the user sees a window open, there's some flickering from lots of retries to launch a custom tab, and then it gives up and fails.

The most prominent case for this is when Android's parental controls are active and they have blocked Chrome (or whichever browser is installed on the device). To make this case clearer to the user, this introduces a dialogue telling the user that their browser is blocked and provides a link to parental controls in settings. There are admittedly some use cases where this can happen for reasons other than parental controls, but this was deemed not worth the complexity since this is a short term solution for a very niche case.

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
@@ -67,9 +67,11 @@ public class TwaLauncher {
      */
     private static final String ACTION_SUPERVISION_SETTINGS = "android.settings.SUPERVISION_SETTINGS";
 
-    /** Strategy for showing a dialog when no browser is available. */
+    /** 
+     * Strategy for showing a dialog when no browser is available. Interface exists just to
+     * facilitate unit testing.
+     */
     public interface BrowserUnavailableDialogStrategy {
-        /** Shows the dialog. */
         void show(Activity activity);
     }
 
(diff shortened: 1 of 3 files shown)


gstepniewski-google (changes requested):

gstepniewski-google (approved):
Approving following offline discussion

gstepniewski-google (approved):

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
```diff
@@ -338,6 +377,39 @@ public void setStartupUptimeMillis(long startupUptimeMillis) {
...
+
+        // Query for all browsers that can handle a standard URL. This allows us to find the
+        // user's preferred browser to provide a helpful name in the dialog.
+        Intent queryBrowsersIntent = new Intent()
+                .setAction(Intent.ACTION_VIEW)
+                .addCategory(Intent.CATEGORY_BROWSABLE)
+                .setData(Uri.fromParts("http", "", null));
+
+        List<ResolveInfo> allBrowsers = pm.queryIntentActivities(queryBrowsersIntent,
+                PackageManager.MATCH_DEFAULT_ONLY | PackageManager.MATCH_UNINSTALLED_PACKAGES);
```
dargmuesli, line 396: Isn't MATCH_UNINSTALLED_PACKAGES only available API level 24 onwards? I think this project is currently set to minSdkVersion 23.

gstepniewski-google: [MATCH_UNINSTALLED_PACKAGES](https://developer.android.com/reference/android/content/pm/PackageManager#MATCH_UNINSTALLED_PACKAGES) was indeed added in API 24, but it was introduced to replace [GET_UNINSTALLED_PACKAGES](https://developer.android.com/reference/android/content/pm/PackageManager#GET_UNINSTALLED_PACKAGES), deprecated at the same time. Both have the same constant value of 8192 (0x00002000), which means this was basically a name change.

PR #429, opened 2023-06-14
Add TWA/CCT postMessage demo

SayedElabady:
Add a demo for using postMessage API in CCT/TWA, the feature is not available on Chrome stable yet.

Issue #526, opened 2025-05-25
SecurityException when binding to PostMessageService in TWA integration


ehsan323:
Description:
I'm integrating a Trusted Web Activity (TWA) into my Android app and trying to enable postMessage communication. However, I'm encountering the following issue:

Crash Log:
java.lang.SecurityException: Not allowed to bind to service Intent { cmp=com.emofid.rnmofid/androidx.browser.customtabs.PostMessageService }

This happens after I added the following to my AndroidManifest.xml:
<service
    android:name="androidx.browser.customtabs.PostMessageService"
    android:exported="true"
    android:permission="android.permission.BIND_CUSTOM_TABS_SERVICE">
    <intent-filter>
        <action android:name="android.support.customtabs.action.CustomTabsService" />
    </intent-filter>
</service>

In my code, this line returns false and the postMessage connection never succeeds:
boolean result = mSession.requestPostMessageChannel(URL);

Dependencies Used:
implementation "androidx.appcompat:appcompat:1.6.1"
implementation "androidx.browser:browser:1.7.0"
implementation "com.google.androidbrowserhelper:androidbrowserhelper:2.5.0"

Expected Behavior:
The call to requestPostMessageChannel() should return true, and the message channel should be established.

Actual Behavior:
The app crashes with a SecurityException related to PostMessageService, or requestPostMessageChannel() returns false when the service is not declared.

Steps to Reproduce:

Integrate TWA using the dependencies above.
Add PostMessageService to the manifest.
Attempt to call requestPostMessageChannel().

Additional Notes:

Digital Asset Links (DAL) verification is confirmed and working.
TWA launches correctly and shows content.
Issue occurs consistently when attempting to establish postMessage communication.

The TWA is created by BubbleCLI.

Issue #514, opened 2025-05-12
Twa postMessage event is not received on js.


MohammadrasiM:
I used Twa postMessage Demo and the channel is created and the postMessage event is sent but I couldn't receive the message By window message event listener. I need help .

MohammadrasiM:
@rafaeldev  hi thx for the suggestion but i need it to work with postMessage  its safer.

Issue #56, opened 2020-01-15
Add a postMessage demo.


PEConn:
Using postMessage is pretty confusing, we should add a demo for it.

(This is for CCTs, not for TWAs)

andreban:
My suggestion is to prefix the demo with `cct`, to keep it consistent with the TWA demos, which are prefixed with `twa`. Eg: `cct-postmessage`.

FluorescentHallucinogen:
@andreban @PEConn Any news on postMessage support in TWA? This is an essential (and one of the most requested) feature. See my rationale here: https://github.com/GoogleChromeLabs/svgomg-twa/issues/84.

andreban:
We don't have an update on postMessage for Trusted Web Activity.

Even though we constantly re-evaluate postMessage for Trusted Web Activity, the overall the goal is to use Web Platform APIs where one is available or planned. Where a capability is not available / planned, we're looking into adding those to the platform. 

The advantage here is that by doing it this way we ensure implementations will work the same way across platforms and browsers and avoid a situation where developers need to have multiple codebases. This does indeed take longer than just adding postMessage but is better in the long-term.

Nevertheless, we're looking for feedback for features and capabilities that might not make sense as Web APIs but would make sense as postMessage.

[Project Fugu](https://goo.gle/fugu-api-tracker) already has issues open for both geofencing and geotracking, the use cases mentioned on the rationale. The Chrome issue for geofencing is [crbug.com/898533 ](https://crbug.com/898533)and the issue for geotracking is [crbug.com/898536](https://crbug.com/898536)

@b1tr0t FYI

FluorescentHallucinogen:
> The advantage here is that by doing it this way we ensure implementations will work the same way across platforms and browsers and avoid a situation where developers need to have multiple codebases. This does indeed take longer than just adding postMessage but is better in the long-term.

TWA is an Android app, so there is only one codebase (native Android code + Web code).

TWA is not only for Chrome. Other browsers like Firefox may implement some Web APIs unacceptably long or even not implement at all, which lead to bad UX and fragmentation (multiple codebases for different browsers, but that use kludges instead of postMessage).

micDropper:
It is now a year later and Web API's are still limiting development in many ways. If it is a _Trusted_ Web Activity, then it seems to me that Android should trust the website postMessage's completely and expose all of Android's API's, a place you (Google) can exert API control outside of Chrome. 

[...]

PR #584, opened 2026-05-01
Expose CustomTabsSession on TwaLauncher and LauncherActivity

### LINKED ISSUE #472: Trusted Web Activity Bubblewrap with PostMessage
**Is your feature request related to a problem? Please describe.** 

I am currently working on a project where I need to manage the sending of post messages to the browser. I have successfully overridden the getCustomTabsCallback method in the LauncherActivity class to return a custom class extended from the QualityEnforcer class. This extended class contains the logic of CustomTabsCallback as described in the [documentation](https://developer.chrome.com/docs/android/post-message-twa?hl=fr#create_customtabscallback). However, to do this, I need access to the session created in the onCustomTabsServiceConnected method of the TwaCustomTabsServiceConnection subclass in the TwaLauncher class.

**Describe the solution you’d like** 

I would like to have a getter method or some other way to access the CustomTabsSession mSession property from the TwaLauncher class. Currently, this property is private and there is no getter method to retrieve it.

**Describe alternatives you’ve considered** 

As an alternative, I've tried overriding the createTwaLauncher method in the LauncherActivity class to return an extended custom class from the TwaLauncher class to access the mSession property of CustomTabsSession, or overriding the TwaCustomTabsServiceConnection subclass. However, they are still private, so this requires modifying the library's source code, which isn't ideal.

**Additional context** 

I have successfully made postal messages work based on the “twa-post-message” demo and the [documentation](https://developer.chrome.com/docs/android/post-message-twa?hl=fr#create_customtabscallback) without any problem where I was able to communicate between the TWA application and my Web app. The thing is, I use Bubblewrap to generate the project and the structure has nothing to do with that of the demo. I would simply like to add a new class to the project generated by Bubblewrap without having to modify the entire project structure and having to maintain it with each new update of the “android-browser-helper” repo.

If it is possible to find a solution to use post messages in a project generated by Bubblewrap without having to make hacks, it would be greatly appreciated.

### LINKED ISSUE #510: Bubblewrap/LauncherActivity will not work with post messages
**Problem**
There is currently no way to use post messages with `LauncherActivity` as used by a Bubblewrap generated app.

The provided demo code uses a completely different structure to launch the TWA and it's not clear how to convert Bubblewrap generated code to use this different structure or exactly what is lost by no longer using `LauncherActivity`/`TwaLauncher`.

**The solution so far**
Here's an implementation that _almost_ works:

```kotlin
package com.example.app

import android.content.ComponentName
import android.content.pm.ActivityInfo
import android.net.Uri
import android.os.Build
import android.os.Bundle
import android.os.Parcel
import android.os.RemoteException
import android.util.Log
import androidx.browser.customtabs.CustomTabsCallback
import androidx.browser.customtabs.CustomTabsClient
import androidx.browser.customtabs.CustomTabsServiceConnection
import androidx.browser.customtabs.CustomTabsSession
import com.google.androidbrowserhelper.trusted.LauncherActivity
import com.google.androidbrowserhelper.trusted.QualityEnforcer
import com.google.androidbrowserhelper.trusted.TwaLauncher

class LauncherActivity : LauncherActivity() {
    private var mSession: CustomTabsSession? = null
    private var mValidated = false
    private var logTag = "Example"
    private var mLauncherActivity: LauncherActivity? = null

    override fun onCreate(savedInstanceState: Bundle?) {
        mLauncherActivity = this

        super.onCreate(savedInstanceState)

        val packageName = CustomTabsClient.getPackageName(this, null)

        CustomTabsClient.bindCustomTabsServicePreservePriority(
            this,
            packageName, // Mora biti Chrome ili browser koji podržava Custom Tabs
            customTabsServiceConnection
        )

        // Setting an orientation crashes the app due to the transparent background on Android 8.0
        // Oreo and below. We only set the orientation on Oreo and above. This only affects the
        // splash screen and Chrome will still respect the orientation.
        // See https://github.com/GoogleChromeLabs/bubblewrap/issues/496 for details.
        requestedOrientation = if (Build.VERSION.SDK_INT > Build.VERSION_CODES.O) {
            ActivityInfo.SCREEN_ORIENTATION_USER_PORTRAIT
        } else {
            ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED
        }
    }

    override fun onDestroy() {
        mSession = null
        mLauncherActivity = null
        baseContext.unbindService(customTabsServiceConnection)

        super.onDestroy()
    }

    private val customTabsServiceConnection = object : CustomTabsServiceConnection() {
        override fun onCustomTabsServiceConnected(name: ComponentName, client: CustomTabsClient) {
            client.warmup(0L)
        }

        override fun onServiceDisconnected(name: ComponentName) {}
    }

    private val customTabsCallback: CustomTabsCallback = object : QualityEnforcer() {
        override fun onPostMessage(message: String, extras: Bundle?) {
            super.onPostMessage(message, extras)

            Log.d(logTag, "Got message: $message")
        }

        override fun onRelationshipValidationResult(
            relation: Int, requestedOrigin: Uri,
            result: Boolean, extras: Bundle?
        ) {
            mValidated = result
        }

        // Listens for navigation, requests the postMessage channel when one completes.
        override fun onNavigationEvent(navigationEvent: Int, extras: Bundle?) {
            // Waiting for navigation finished _should_ mean that the pwa is running before we
            // continue.
            if (navigationEvent != NAVIGATION_FINISHED) {
                return
            }

            //
            // Here we use java reflection to get the mSession from mTwaLauncher, as set in the
            // parent LauncherActivity class.
            //
            // This is very much beyond being a bad idea. If you're using reflection,
            // you're probably doing it wrong.
            //
            if (mLauncherActivity != null) {
                val twaLauncherField =
                    mLauncherActivity?.javaClass?.superclass?.getDeclaredField("mTwaLauncher")
                twaLauncherField?.isAccessible = true
                val mTwaLauncher = twaLauncherField?.get(mLauncherActivity) as? TwaLauncher

                if (mTwaLauncher != null) {
                    val sessionField = mTwaLauncher?.javaClass?.getDeclaredField("mSession")
                    sessionField?.isAccessible = true
                    mSession = sessionField?.get(mTwaLauncher) as? CustomTabsSession
                }
            }

            if (mSession == null) {
                return
            }
            //
            // End reflection nonsense.
            //

            val result = mSession!!.requestPostMessageChannel(getLaunchingUrl())

            Log.d(logTag, "Requested Post Message Channel: $result")
        }

        override fun onMessageChannelReady(extras: Bundle?) {
            Log.d(logTag, "Message channel ready.")

            if (mSession == null) {
                return
            }

            mSession.postMessage("hello", null)
        }
    }

    override fun getCustomTabsCallback(): CustomTabsCallback {
        return customTabsCallback
    }
}
```

`customTabsCallback` inherits from `QualityEnforcer` in order to retain existing behaviour and `getCustomCallback` is overridden to return our custom tabs callback object. 

On it's own, this should be enough as warmup is already called on the client object in `TwaLauncher` however if we leave it at that we never get navigation events. To receive those we also have to call `CustomTabsClient.bindCustomTabsService(customTabsServiceConnection)` which calls warmup again and absolutely nothing else. With that in place we receive navigation events and are able to send and receive post messages.

Among the many problems, and they are many, are that it requires reflection to pull the `mSession` value from `TwaLauncher` and that it also errors out with the following whenever the app is dismissed:

```
2025-05-06 13:42:02.117  4383-4394  IPCThreadState          com.example.app                   E  Binder transaction failure. id: 124609, BR_*: 29189, error: -22 (Invalid argument)
2025-05-06 13:42:02.118  4383-4394  JavaBinder              com.example.app                   E  *** Uncaught remote exception! Exceptions are not yet supported across processes. Client PID 2511 UID 10146. (Ask Gemini)
  java.lang.RuntimeException: android.os.DeadObjectException
  at android.os.Parcel.writeException(Parcel.java:3012)
  at android.os.Binder.execTransactInternal(Binder.java:1524)
  at android.os.Binder.execTransact(Binder.java:1444)
Caused by: android.os.DeadObjectException
  at android.os.BinderProxy.transactNative(Native Method)
  at android.os.BinderProxy.transact(BinderProxy.java:586)
  at android.support.customtabs.ICustomTabsCallback$Stub$Proxy.onMessageChannelReady(ICustomTabsCallback.java:259)
  at androidx.browser.customtabs.PostMessageService$1.onMessageChannelReady(PostMessageService.java:40)
  at android.support.customtabs.IPostMessageService$Stub.onTransact(IPostMessageService.java:70)
  at android.os.Binder.execTransactInternal(Binder.java:1500)
  at android.os.Binder.execTransact(Binder.java:1444) 
```

`onDestroy` on `LauncherActivity` is called and then, sometime later, a `TAB_HIDDEN` navigation event is passed to `CustomTabsCallback.onNavigationEvent` and then the preceding error is thrown. Once thrown, the app is unable to send or receive post messages on any future launch.

I have not been able to find any way to unbind services or callbacks to avoid this happening.

**Solutions**

Ideally, someone who knows better will look at this and point out the obvious thing I'm missing that stops `DeadObjectException` from being thrown.

Failing that, perhaps someone can write a demo showing _how_ to integrate post messages with Bubblewrap/`LauncherActivity`.

dnikolaev:
## Summary
Adds two `@Nullable` read-only accessors over fields that already exist:

- `TwaLauncher#getSession()` (public)
- `LauncherActivity#getCustomTabsSession()` (protected)

No behavior change for callers that don't use them.

## Motivation
Apps that subclass `LauncherActivity` (e.g. anything generated by Bubblewrap) can't reach the `CustomTabsSession` that `TwaLauncher` constructs internally. Since Chrome 115 enabled `postMessage` in TWAs, that session is the only entry point to `requestPostMessageChannel` / `postMessage`, so the high-level launcher is currently incompatible with PostMessage.

The `demos/twa-post-message` sample added in #429 builds its own session and does not extend `LauncherActivity`, so it doesn't translate to the Bubblewrap launcher path. In the wild this is worked around with reflection on `mSession` or by spinning up a parallel `CustomTabsServiceConnection` and relaunching the TWA — both fragile.

## Lifecycle
The session is established asynchronously in `onCustomTabsServiceConnected` and cleared in `onServiceDisconnected` / `destroy()`. Both getters are `@Nullable` and the Javadoc tells callers to guard.

## Usage
With this in place, a subclass can wire PostMessage on the existing `getCustomTabsCallback()` hook:

```java
@Override
protected CustomTabsCallback getCustomTabsCallback() {
    return new CustomTabsCallback() {
        @Override public void onNavigationEvent(int event, Bundle extras) {
            if (event != NAVIGATION_FINISHED) return;
            CustomTabsSession s = getCustomTabsSession();
            if (s != null) s.requestPostMessageChannel(source, target, new Bundle());
        }
        @Override public void onPostMessage(@NonNull String msg, @Nullable Bundle extras) {
            CustomTabsSession s = getCustomTabsSession();
            if (s != null) s.postMessage("pong", null);
        }
    };
}
```

Verified end-to-end in a Bubblewrap-generated production app.

## Scope
- No `postMessage` helpers or callback wrappers — those belong in `androidx.browser.customtabs`.
- No new fields, lifecycle changes, threads, or dependencies.
- No change to session creation timing or `DEFAULT_SESSION_ID`.

Closes #472. Closes #510. Refs GoogleChromeLabs/bubblewrap#954.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
@@ -30,6 +30,7 @@ import androidx.annotation.Nullable;
 import androidx.browser.customtabs.CustomTabColorSchemeParams;
 import androidx.browser.customtabs.CustomTabsCallback;
 import androidx.browser.customtabs.CustomTabsIntent;
+import androidx.browser.customtabs.CustomTabsSession;
 import androidx.browser.trusted.FileHandlingData;
 import androidx.browser.trusted.TrustedWebActivityDisplayMode;
 import androidx.browser.trusted.TrustedWebActivityIntentBuilder;
@@ -278,6 +279,19 @@ public class LauncherActivity extends Activity {
     protected TwaLauncher createTwaLauncher() {
         return new TwaLauncher(this, mMetadata.launchingBrowser, SessionStore.makeSessionId(getTaskId()),
                 new SharedPreferencesTokenStore(this));
+    }
+
+    /**
+     * Returns the current CustomTabsSession from the TwaLauncher.
+     *
+     * The session becomes available after the TWA is launched and the service connects.
+     * Can be used for advanced features like PostMessage.
+     *
+     * @return The active CustomTabsSession, or null if TWA hasn't launched yet or service disconnected
+     */
+    @Nullable
+    protected CustomTabsSession getCustomTabsSession() {
+        return mTwaLauncher != null ? mTwaLauncher.getSession() : null;
     }
 
     private boolean splashScreenNeeded() {
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivity.java
@@ -30,6 +30,7 @@ import androidx.annotation.Nullable;
 import androidx.browser.customtabs.CustomTabColorSchemeParams;
 import androidx.browser.customtabs.CustomTabsCallback;
 import androidx.browser.customtabs.CustomTabsIntent;
+import androidx.browser.customtabs.CustomTabsSession;
 import androidx.browser.trusted.FileHandlingData;
 import androidx.browser.trusted.TrustedWebActivityDisplayMode;
 import androidx.browser.trusted.TrustedWebActivityIntentBuilder;
@@ -280,6 +281,19 @@ public class LauncherActivity extends Activity {
     protected TwaLauncher createTwaLauncher() {
         return new TwaLauncher(this, mMetadata.launchingBrowser, SessionStore.makeSessionId(getTaskId()),
                 new SharedPreferencesTokenStore(this));
+    }
+
+    /**
+     * Returns the current CustomTabsSession from the TwaLauncher.
+     *
+     * The session becomes available after the TWA is launched and the service connects.
+     * Can be used for advanced features like PostMessage.
+     *
+     * @return The active CustomTabsSession, or null if TWA hasn't launched yet or service disconnected
+     */
+    @Nullable
+    protected CustomTabsSession getCustomTabsSession() {
+        return mTwaLauncher != null ? mTwaLauncher.getSession() : null;
     }
 
     private boolean splashScreenNeeded() {
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
@@ -389,6 +389,20 @@ public class TwaLauncher {
     }
 
     /**
+     * Returns the current CustomTabsSession, if available.
+     *
+     * This is useful for advanced use cases like PostMessage that require direct
+     * session access. The session becomes available after the service connects and
+     * may become null if the service disconnects.
+     *
+     * @return The active CustomTabsSession, or null if not yet established or disconnected
+     */
+    @Nullable
+    public CustomTabsSession getSession() {
+        return mSession;
+    }
+
+    /**
      * Sets the timestamp (in SystemClock.uptimeMillis()) when the TWA launcher
      * activity was created. This timestamp is used to report the full startup
      * duration to the browser.
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/TwaLauncher.java
@@ -401,6 +401,20 @@ public class TwaLauncher {
     }
 
     /**
+     * Returns the current CustomTabsSession, if available.
+     *
+     * This is useful for advanced use cases like PostMessage that require direct
+     * session access. The session becomes available after the service connects and
+     * may become null if the service disconnects.
+     *
+     * @return The active CustomTabsSession, or null if not yet established or disconnected
+     */
+    @Nullable
+    public CustomTabsSession getSession() {
+        return mSession;
+    }
+
+    /**
      * Sets the timestamp (in SystemClock.uptimeMillis()) when the TWA launcher
      * activity was created. This timestamp is used to report the full startup
      * duration to the browser.


dnikolaev:
@googlebot I've signed CLA.

gstepniewski-google:
Hi Dima, thanks for submitting this! I have two comments:

1. I like the idea of exposing the CustomTabsSession, but I'm not fully sold on the implementation - you mention in the javadoc that the field is initially null and then becomes available when session is established. Would it be a good idea to return the session wrapped in an observable so that clients can subscribe to it and automatically perform actions when it becomes available? There's a `launchWhenSessionEstablished` in `TwaLauncher` that could be used for this.

2. The CLA check is still failing despite me rerunning it a couple of times - do you mind double checking that everything is correct on your end?

Thanks!

sergio-veed:
Would really love to see this merged if possible, it would simplify the initialisation code in our app quite a bit!

raffij:
@dnikolaev is this something you are wanting to finish off yourself?

dnikolaev:
@gstepniewski-google Thanks for the review, and apologies for the slow follow-up — both points are now addressed or actionable:

**1. CLA — fixed.** The check was failing on a `Co-Authored-By` trailer pointing at an AI assistant's no-reply address, which can never sign the CLA (my own signature was recognized all along). I've rewritten the commit to drop the trailer and rebased onto current `main`; the PR is now a single commit with the same 2-file / +28 diff, and both `cla/google` and `check-changes` are green.

**2. The observable question.** You're right that the async lifecycle is the sharp edge — and it got sharper with #549 (LauncherActivity now finishes after launch), so I'd rather converge on a shape you'd accept than defend the minimal getter. Why I started with plain accessors:

- Every user in #472 / #56 / #510 / #514 / #526 needed *access* and built something worse to get it: reflection into `mSession`, a parallel `CustomTabsServiceConnection` with a duplicate activity, or copying `LauncherActivity`+`TwaLauncher` wholesale into their app. A nullable getter directly replaces those.
- `launchWhenSessionEstablished` is private and single-shot; exposing the session through an observable means designing a new public surface (threading, multiple subscribers, behavior on disconnect/reconnect) — a bigger API commitment than this PR intends.
- The getters follow the existing pattern of nullable accessors and document the lifecycle in the javadoc.

That said, if you'd prefer a callback, I'm happy to add one in this PR — two shapes that stay small:

a) a protected hook `LauncherActivity#onCustomTabsSessionAvailable(CustomTabsSession)` invoked when the connection delivers the session; or
b) a registration API on `TwaLauncher`: `addOnSessionAvailableListener(Consumer<CustomTabsSession>)`, delivered immediately if the session already exists.

Tell me which (either, both, or getter-only) and I'll push it right away.

One scoping note: the channel-reliability problems around postMessage (#510's timing flakiness, the Chrome 150 regression in #602) are Chromium-side and orthogonal — this change is only about reaching the session that already exists.

@raffij yes — finishing this myself; today's rebase + CLA fix are that, and I'll turn the API decision around as soon as there's a signal. @sergio-veed thanks for the support — hoping this gets us over the line.

gstepniewski-google:
Hi Dima, thanks for working on this!

I agree that the full observable pattern is a bit of an overkill here and we should try to make keep this change smaller.
I like your idea of adding a protected hook to LauncherActivity, it's very simple, to the point, and supports clients who follow this approach (including bubblewrap).
For TwaLauncher, I think we might be able to leverage the already existing completionCallback in launch(): since session is guaranteed to be non-null by the time it fires, calling the getter inside it should be a safe operation - which means we don't need the addListener method (and all the callback management boilerplate that would have to follow).

Thanks!

PR #598, opened 2026-07-13
Refactor queryPurchases to use QueryPurchasesParams

SharkMan201:
Refactored the `PlayBillingWrapper.queryPurchases` method to use `QueryPurchasesParams` instead of a plain String parameter when calling `BillingClient.queryPurchasesAsync`, matching the new Play Billing Library v8 API specification.


gstepniewski-google (approved):

jenn021998-cloud:
please cancel its all done

PR #600, opened 2026-07-13
Make listPurchaseHistory return error/empty list immediately and remove queryPurchaseHistory

SharkMan201:
Updated `ListPurchaseHistoryCall` to immediately return `CHROMIUM_RESULT_ERROR` and an empty list when called, completely bypassing the Play Billing Client connection and queries, as `listPurchaseHistory` is no longer supported in PBL v8. Completely removed the obsolete `queryPurchaseHistory` method and its unused imports from the `BillingWrapper` interface and all of its implementations (`PlayBillingWrapper`, `MockBillingWrapper`, and `ConnectedBillingWrapper`). Updated `ListPurchaseHistoryCallTest` to verify that an error and an empty list are returned immediately without any Billing Client interaction.

### DIFF
--- a/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/ListPurchaseHistoryCall.java
+++ b/playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/ListPurchaseHistoryCall.java
@@ -19,13 +19,7 @@ import android.os.Parcelable;
 
 import androidx.annotation.Nullable;
 
-import com.android.billingclient.api.BillingClient;
-import com.android.billingclient.api.BillingResult;
-import com.android.billingclient.api.Purchase;
-import com.android.billingclient.api.PurchaseHistoryRecord;
 import com.google.androidbrowserhelper.playbilling.provider.BillingWrapper;
-
-import java.util.List;
 
 /**
  * A class for parsing Digital Goods API calls from the browser and converting them into a format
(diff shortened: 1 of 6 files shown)


playbilling/src/main/java/com/google/androidbrowserhelper/playbilling/digitalgoods/ListPurchaseHistoryCall.java
```diff
@@ -56,29 +50,9 @@ public static ListPurchaseHistoryCall create(@Nullable DigitalGoodsCallback call
...
-            int index = 0;
-            for (PurchaseHistoryRecord record : purchaseHistoryList) {
-                parcelables[index++] = PurchaseDetails.create(record).toBundle();
-            }
-        }
-
         Bundle args = new Bundle();
-        args.putInt(KEY_RESPONSE_CODE, DigitalGoodsConverter.toChromiumResponseCode(result));
-        args.putParcelableArray(KEY_PURCHASE_HISTORY_LIST, parcelables);
+        args.putInt(KEY_RESPONSE_CODE, DigitalGoodsConverter.CHROMIUM_RESULT_ERROR);
```
gstepniewski-google: I think we should return `CHROMIUM_RESULT_OK` here instead (otherwise this will result in rejected promises instead of 'quiet failure')

gstepniewski-google (changes requested):

gstepniewski-google (approved):

PR #607, opened 2026-08-06
Support USE_HIGH_PRI_NOTIFICATIONS manifest metadata for sending high priority notifications for TWA

Dp-Goog:
Google internal: Please see b/491724839 for more information.

This produces a new metadata tag called `USE_HIGH_PRI_NOTIFICATIONS` that can be used to have TWAs use notifications in channels of higher priority, instead of the default priority.

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/NotificationUtils.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/NotificationUtils.java
@@ -14,13 +14,11 @@
 
 package com.google.androidbrowserhelper.trusted;
 
-import android.app.Activity;
 import android.app.NotificationChannel;
 import android.app.NotificationManager;
 import android.app.Service;
 import android.content.ComponentName;
 import android.content.Context;
-import android.content.pm.ActivityInfo;
 import android.content.pm.PackageManager;
 import android.content.pm.ServiceInfo;
 import android.os.Build;
@@ -34,12 +32,6 @@ import java.util.Locale;
 public class NotificationUtils {
     /**
      * Metadata key to specify whether notification channels should be created with high importance.
-     */
-    public static final String METADATA_USE_HIGH_PRI_NOTIFICATIONS =
-            "android.support.customtabs.trusted.USE_HIGH_PRI_NOTIFICATIONS";
-
-    /**
-     * AndroidX alias for {@link #METADATA_USE_HIGH_PRI_NOTIFICATIONS}.
      */
     public static final String METADATA_USE_HIGH_PRI_NOTIFICATIONS_ANDROIDX =
             "androidx.browser.trusted.USE_HIGH_PRI_NOTIFICATIONS";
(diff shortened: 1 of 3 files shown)


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/NotificationUtils.java
```diff
@@ -41,21 +60,95 @@ public static boolean areNotificationsEnabled(Context context, String channelNam
     }
 
     /**
-     * Creates a notification channel using the given channel name.
+     * Checks if high-priority notifications are configured in the manifest metadata.
+     */
+    public static boolean shouldUseHighPriorityNotifications(Context context) {
+        if (context == null) return false;
```
PEConn: context should never be null, so we can remove this.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/NotificationUtils.java
```diff
@@ -14,17 +14,36 @@
...
+    /**
+     * Metadata key to specify whether notification channels should be created with high importance.
+     */
+    public static final String METADATA_USE_HIGH_PRI_NOTIFICATIONS =
+            "android.support.customtabs.trusted.USE_HIGH_PRI_NOTIFICATIONS";
+
+    /**
+     * AndroidX alias for {@link #METADATA_USE_HIGH_PRI_NOTIFICATIONS}.
+     */
+    public static final String METADATA_USE_HIGH_PRI_NOTIFICATIONS_ANDROIDX =
```
PEConn, line 37: Why do we need two different constants?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/NotificationUtils.java
```diff
@@ -41,21 +60,95 @@ public static boolean areNotificationsEnabled(Context context, String channelNam
     }
 
     /**
-     * Creates a notification channel using the given channel name.
+     * Checks if high-priority notifications are configured in the manifest metadata.
+     */
+    public static boolean shouldUseHighPriorityNotifications(Context context) {
+        if (context == null) return false;
+        try {
+            Bundle metaData = context.getPackageManager().getApplicationInfo(
```
PEConn: IIUC, you're checking for whether the metadata is in the application section, the Activity section or the Service section? We should just put it in a single section.

PEConn (commented):
I think the general idea is good, but we can make the code a lot simpler if we're more prescriptive - eg, saving the metadata has to be in one specific place, in one specific format.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/NotificationUtils.java
Dp-Goog: Implemented.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/NotificationUtils.java
Dp-Goog: The first one was for [TWAs that hadn't migrated to AndroidX yet](https://developer.android.com/jetpack/androidx/migrate) who wanted to use. I kept the `AndroidX` version because it seems like that is the latest. Does that sound good?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/NotificationUtils.java
Dp-Goog: I put in the Service section, because the `NotificationDelegationService` kinda uses it. Please let me know if that is a reasonable place, or somewhere else might be better.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/NotificationUtils.java
PEConn: SGTM, thanks

PEConn (approved):
LGTM, thanks

Dp-Goog:
@gstepniewski-google can you help approve this? Merging is blocked for me, because "At least 1 approving review is required by reviewers with write access.".

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/NotificationUtils.java
```diff
@@ -41,21 +52,82 @@ public static boolean areNotificationsEnabled(Context context, String channelNam
...
+                channelName, importance);
         NotificationManagerCompat.from(context).createNotificationChannel(channel);
     }
 
     /**
      * Generates a notification channel id from a channel name.
      * TODO: Remove this when we can use the method defined in AndroidX instead.
      */
-    private static String channelNameToId(String name) {
+    public static String channelNameToId(String name) {
```
gdziadosz-google: Is it necessary to make this method public?
Same with `shouldUseHighPriorityNotifications`.

Dp-Goog: I initially made them public because both of them were being used in `NotificationUtilsTest`, but I have now removed their acccessibility and added a `@VisibleForTesting` so that they now have package-private visibility (AKA default access), but is still available for testing. That way it is not publicly accessible outside the package.

androidbrowserhelper/src/test/java/com/google/androidbrowserhelper/trusted/NotificationUtilsTest.java
```diff
@@ -0,0 +1,260 @@
...
+
+package com.google.androidbrowserhelper.trusted;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.robolectric.Shadows.shadowOf;
+
+import android.app.Activity;
```
dtonderski-google: I think we have a few unused imports here, could we clean that up?

androidbrowserhelper/src/test/java/com/google/androidbrowserhelper/trusted/NotificationUtilsTest.java
```diff
@@ -0,0 +1,260 @@
...
+        assertEquals("general_notifications_channel_id",
+                NotificationUtils.channelNameToId("General Notifications"));
+        assertEquals("chat_channel_id",
+                NotificationUtils.channelNameToId("Chat"));
+        assertEquals("test_channel_name_channel_id",
+                NotificationUtils.channelNameToId("Test Channel Name"));
+    }
+
+    @Test
+    public void getNotificationImportance_matchesExpectedImportance() {
```
dtonderski-google: Not a huge fan of this approach - the entire class is parametrized, but only three tests actually use the parameters. IIUC, this means that the remaining tests are unnecessarily run seven times.

WDYT about extracting the tests that specifically use the parametrized values into a separate parametrized test class, and make `NotificationUtilsTest` non-parametrized? This is similar to the approach of [ToPriceTest](https://github.com/GoogleChrome/android-browser-helper/blob/a3638f23537189f82165ff96fb2a60431b03f34c/playbilling/src/test/java/com/google/androidbrowserhelper/playbilling/digitalgoods/ToPriceTest.java).

dtonderski-google (changes requested):

androidbrowserhelper/src/test/java/com/google/androidbrowserhelper/trusted/NotificationUtilsTest.java
Dp-Goog: Done.

androidbrowserhelper/src/test/java/com/google/androidbrowserhelper/trusted/NotificationUtilsTest.java
Dp-Goog: Done, how does it look now?

gdziadosz-google (approved):

dtonderski-google (approved):

PR #610, opened 2026-08-12
Adding AI harness with agent definitions, skills, and rules

dmurph:
This is an MVP with the structure, the documentation can be edited as we find AI to miss stuff or be over-prescriptive, etc.


dmurph:
moving to draft, needs work

PR #608, opened 2026-08-06
Add TWA high priority notifications test site

Dp-Goog:
Create a sample test TWA that wraps https://dp-goog.github.io/notifications-api-testing/, so that SW notifications can be sent. This mostly uses the artifacts used in `twa-notification-delegation`, with the code changes being:

- The new `androidx.browser.trusted.USE_HIGH_PRI_NOTIFICATIONS` field set.
- The TWA launches `https://dp-goog.github.io/notifications-api-testing/` instead.
- `onCreate()` and `onDestroy()` logs for notifications

Verified that it works, based on the following logs:

Screenshot 2026-08-06 at 4 37 27 PM

PR #612, opened 2026-08-13
Migrate notification channels for TWAs to reuse high priority if already installed

Dp-Goog:
If a TWA is already installed, then just changing the priority of the notifications will not "update" the existing notification channels, even if the manifest field is set.

This PR fixes that by doing a "migration", by ensuring that if `androidx.browser.trusted.USE_HIGH_PRI_NOTIFICATIONS` is set to true for existing TWAs, a new channelId is used, which retriggers notifications permissions and notification channel
creation. This helps fully solve #607 .

### DIFF
--- a/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/NotificationUtils.java
+++ b/androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/NotificationUtils.java
@@ -48,7 +48,7 @@ public class NotificationUtils {
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.O) return true;
 
         NotificationChannel channel =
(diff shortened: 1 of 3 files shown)


Dp-Goog:
@PEConn or @dtonderski-google or @gdziadosz-google PTAL at this, thanks!

PEConn (approved):
This seems reasonable - although I do suggest you test it manually - both when the previous channel existed and when it didn't.

Dp-Goog:
That's a good idea, I just tried out the changes with the test site from #608 , and this seems to be working.

PEConn (approved):

glennhartmann (approved):
I don't have a ton of expertise in this area, but it looks reasonable to me, and I trust @PEConn 's judgment. rslgtm

PR #613, opened 2026-08-13
Add ShortcutTrampolineActivity for TWA shortcuts

dmurph:
### Summary

Introduces `ShortcutTrampolineActivity`, a lightweight, dedicated trampoline activity designed specifically for TWA shortcuts (`shortcuts.xml`). It resolves the window unresponsiveness and taskbar disappearance bug on Android Desktop by adhering to `android:noDisplay="true"` requirements and delegating launch handling asynchronously via Application Context.

---

### Problem & Background

When TWA shortcuts target `LauncherActivity`, Android Desktop's WindowManager detects a translucent activity entering the TWA task stack and initiates a desktop-to-fullscreen window transition. When `LauncherActivity` finishes quickly thereafter, WindowManager's window state desynchronizes from the inner activity bounds, resulting in:
- An unresponsive TWA window (mouse/keyboard input blocked).
- Missing taskbar icon and broken desktop window management.

Because `LauncherActivity` performs asynchronous Custom Tabs service binding and may display UI dialogs, it cannot simply be annotated with `android:noDisplay="true"`.

---

### Solution

1. **`ShortcutTrampolineActivity`**:
   - Declared with `android:theme="@android:style/Theme.NoDisplay"` and `android:noDisplay="true"`.
   - Calls `finish()` synchronously in `onCreate()`, satisfying WindowManager constraints and disappearing immediately before any window hierarchy or transition issues occur.

2. **Decoupled Asynchronous Launch via Application Context**:
   - Uses `getApplicationContext()` with `TwaLauncher` so Custom Tabs Service connection, warm-up, and intent launching proceed safely in the background independently of the activity lifecycle.
   - Automatically attaches `FLAG_ACTIVITY_NEW_TASK` to prevent runtime exceptions when launching from a non-Activity context.

3. **Origin & Security Validation**:
   - Validates incoming shortcut URIs against `defaultUrl` and `additionalTrustedOrigins` defined in `LauncherActivityMetadata` to prevent arbitrary or untrusted URL launches.

4. **Documentation & Demo Updates**:
   - Updated `README.md` with instructions on wiring `ShortcutTrampolineActivity` into `shortcuts.xml`.
   - Updated `demos/twa-basic` with example `shortcuts.xml` and manifest integration.

---

### Alternatives Considered

- **Background Service Trampoline:** Rejected due to Android 10+ Background Activity Launch (BAL) restrictions blocking background activity launches when the app is dormant.
- **Routing through `LauncherActivity`:** Rejected because `LauncherActivity`'s translucent presence in the stack still triggers the WindowManager state corruption.
- **Jetpack Splash Screen Rewrite:** Too high risk for this targeted fix, as removing custom splash screen infrastructure would cause visual flash/blank loading regression while web content loads in Chrome. Future improvement potential though.

---

### Testing

- **Unit Tests:** Added unit tests in `ShortcutTrampolineActivityTest` verifying:
  - Synchronous `finish()` in `onCreate()`.
  - Origin verification against trusted origins.
  - Safe application context usage and intent flags.
- **Manual Testing:** Verified on Android Desktop (ARC/ChromeOS) using `demos/twa-basic` with shortcuts. Confirmed window remains responsive in freeform mode and taskbar icon remains active upon shortcut launch.


androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ShortcutTrampolineActivity.java
```diff
@@ -0,0 +1,155 @@
...
+                    return super.onPrepareIntent(intent);
+                }
+            };
+
+            TrustedWebActivityIntentBuilder builder = new TrustedWebActivityIntentBuilder(uri);
+            twaLauncher.launch(
+                    builder,
+                    new QualityEnforcer(),
+                    null /* splashScreenStrategy */,
+                    () -> new android.os.Handler(android.os.Looper.getMainLooper()).post(twaLauncher::destroy),
```
gstepniewski-google: This can cause crashes if service binding fails for any reason (because TwaLauncher.destroy() will call `mContext.unbindService(mServiceConnection)` resulting in `IllegalArgumentException`. 

To fix this, we need to update TwaLauncher to make sure the service was actually bound and only attempt to clean it up if it had.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ShortcutTrampolineActivity.java
```diff
@@ -0,0 +1,155 @@
...
+            if (intent == null) {
+                return;
+            }
+
+            Uri uri = intent.getData();
+            if (uri == null) {
+                return;
+            }
+
+            LauncherActivityMetadata metadata = LauncherActivityMetadata.parse(this);
```
gstepniewski-google, line 59: It seems like some parameters parsed out of the manifest are never used in this flow:

- additionalTrustedOrigins
- displayMode
- screenOrientation
- launchHandlerClientMode
- theme colors

Is this on purpose?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivityMetadata.java
```diff
@@ -371,6 +371,25 @@ public static LauncherActivityMetadata parse(Context context) {
...
             // installed - so should never happen.
         }
+
+        if (!metaData.containsKey(METADATA_DEFAULT_URL)) {
+            try {
+                PackageManager pm = context.getPackageManager();
+                android.content.pm.PackageInfo packageInfo = pm.getPackageInfo(context.getPackageName(),
+                        PackageManager.GET_ACTIVITIES | PackageManager.GET_META_DATA);
+                if (packageInfo.activities != null) {
+                    for (ActivityInfo activityInfo : packageInfo.activities) {
```
gstepniewski-google, line 442: `packageInfo.activities` only includes `<activity>` tags, skipping all `<activity-alias>` entries. 

This could cause all shortcut intents to be silently dropped when an app attaches its metadata to an alias.
It can be fixed by calling `pm.queryIntentActivities()` instead.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ShortcutTrampolineActivity.java
```diff
@@ -0,0 +1,155 @@
...
+                }
+            };
+
+            TrustedWebActivityIntentBuilder builder = new TrustedWebActivityIntentBuilder(uri);
+            twaLauncher.launch(
+                    builder,
+                    new QualityEnforcer(),
+                    null /* splashScreenStrategy */,
+                    () -> new android.os.Handler(android.os.Looper.getMainLooper()).post(twaLauncher::destroy),
+                    new TwaLauncher.FallbackStrategy() {
```
gstepniewski-google, line 85: This fallback implementation seems to ignore whatever is specified by the app metadata (e.g. `android.support.customtabs.trusted.FALLBACK_STRATEGY = "webview"`) and does not check if there are any browsers installed/enabled on the device, leading to a potential crash by `customTabsIntent.launchUrl`.

androidbrowserhelper/src/test/java/com/google/androidbrowserhelper/trusted/ShortcutTrampolineActivityTest.java
```diff
@@ -0,0 +1,132 @@
...
+        // The fallback strategy uses the application context to start the intent, which
+        // should be registered in the shadow application.
+        Intent launchedIntent = shadowOf(RuntimeEnvironment.application).getNextStartedActivity();
+        assertNotNull(launchedIntent);
+        assertEquals(Intent.ACTION_VIEW, launchedIntent.getAction());
+        assertEquals(trustedUri, launchedIntent.getData());
+        
+        // Ensure FLAG_ACTIVITY_NEW_TASK is attached
+        int flags = launchedIntent.getFlags();
+        assertEquals(Intent.FLAG_ACTIVITY_NEW_TASK, flags & Intent.FLAG_ACTIVITY_NEW_TASK);
```
gstepniewski-google, line 124: We should also call `shadowOf(Looper.getMainLooper()).idle()` here to make sure that all callback posted to `Looper.getMainLooper()` are actually ran (Roboelectric does not execute them synchronously by default, the looper must be manually flushed).

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ShortcutTrampolineActivity.java
```diff
@@ -0,0 +1,155 @@
...
+import android.content.Intent;
+import android.net.Uri;
+import android.os.Bundle;
+import android.util.Log;
+
+import androidx.annotation.Nullable;
+import androidx.browser.trusted.TrustedWebActivityIntent;
+import androidx.browser.trusted.TrustedWebActivityIntentBuilder;
+
+import java.util.Objects;
```
gstepniewski-google: nit: unused

gstepniewski-google (changes requested):

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ShortcutTrampolineActivity.java
Dp-Goog: Removed.

androidbrowserhelper/src/test/java/com/google/androidbrowserhelper/trusted/ShortcutTrampolineActivityTest.java
Dp-Goog: Done.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ShortcutTrampolineActivity.java
Dp-Goog: Implemented fallback.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/LauncherActivityMetadata.java
Dp-Goog: There are some subtle nuances here:

**Scenario A: Metadata is on the Target Activity, Launcher filter is on the Alias**

```
<activity android:name=".LauncherActivity">
    <meta-data android:name="DEFAULT_URL" android:value="https://example.com" />
</activity>
<activity-alias android:name=".LauncherAlias" android:targetActivity=".LauncherActivity">
    <intent-filter>
        <action android:name="android.intent.action.MAIN" />
        <category android:name="android.intent.category.LAUNCHER" />
    </intent-filter>
</activity-alias>
```

With getPackageInfo (Original): It finds LauncherActivity in packageInfo.activities, extracts the metadata, and succeeds.
With queryIntentActivities: It resolves .LauncherAlias (because it has the filter), but since the metadata is on .LauncherActivity (which doesn't have the filter, and isn't returned), it scans .LauncherAlias's empty metadata and fails.

**Scenario B: Metadata is on the Alias, Launcher filter is on the Alias**

```
<activity android:name=".LauncherActivity" />
<activity-alias android:name=".LauncherAlias" android:targetActivity=".LauncherActivity">
    <meta-data android:name="DEFAULT_URL" android:value="https://example.com" />
    <intent-filter>
        <action android:name="android.intent.action.MAIN" />
        <category android:name="android.intent.category.LAUNCHER" />
    </intent-filter>
</activity-alias>
```

With getPackageInfo (Original): It scans .LauncherActivity, finds no metadata, and fails (since aliases are ignored).
With queryIntentActivities: It resolves .LauncherAlias with the launcher filter, finds the metadata on it, and succeeds.

Bottom line is I think we need both, so I implemented a fallback based behavior. PTAL if that makes sense.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ShortcutTrampolineActivity.java
Dp-Goog: Fixed, thanks for pointing this out.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/ShortcutTrampolineActivity.java
Dp-Goog: Thanks for pointing these out. I believe we need all the fields other than the following:

Manifest Metadata Key | Field Name in Java | Status | Rationale
--  | -- | --
android.support.customtabs.trusted.METADATA_SHARE_TARGET  | Not Required | Used for receiving shared content from other apps. Static shortcuts cannot act as target endpoints for share intents.
android.support.customtabs.trusted.FILE_HANDLING_ACTION_URL | Not Required | Used for handling file open intents when registered as an Android File Handler. Shortcuts do not open local files.
android.support.customtabs.trusted.SPLASH_IMAGE_DRAWABLE  | Not Required | Splash screen image. Trampolines must finish synchronously without creating a window, so they cannot render splash screens.
android.support.customtabs.trusted.SPLASH_SCREEN_BACKGROUND_COLOR | Not Required | Splash screen background color. (See above).
android.support.customtabs.trusted.SPLASH_SCREEN_FADE_OUT_DURATION | Not Required | Splash screen transition duration. (See above).
android.support.customtabs.trusted.FILE_PROVIDER_AUTHORITY | Not Required | File provider authority used to share splash images with Chrome. (See above).
android.support.customtabs.trusted.START_CHROME_BEFORE_ANIMATION_COMPLETE | Not Required | Timing flag for starting Chrome during the splash screen transition. (See above).

I fixed it by introducing `configureIntentBuilder(TrustedWebActivityIntentBuilder builder, Context context)` inside `LauncherActivityMetadata.java` that sets the required metadata (aside from the ones above) for the TWA activity.

Dp-Goog:
@gstepniewski-google I took over Dan's CL. Is there any way to make me the "owner" of the CL moving forward? I can't seem to assign reviewers.

dmurph:
> @gstepniewski-google I took over Dan's CL. Is there any way to make me the "owner" of the CL moving forward? I can't seem to assign reviewers.

Hm... I can't figure out how to. google says it's not possible, so you'll have to do your own fork and make a new pull request. sorry!

Dp-Goog:
> > @gstepniewski-google I took over Dan's CL. Is there any way to make me the "owner" of the CL moving forward? I can't seem to assign reviewers.
> 
> Hm... I can't figure out how to. google says it's not possible, so you'll have to do your own fork and make a new pull request. sorry!

Well that sucks. I cc'd myself on the internal bug tracking this PR so that I can get updates there. I think we can continue making changes on top of this one, and Greg can help land this.

gstepniewski-google (approved):

PR #615, opened 2026-08-18
Update DelegationService to override default notification channel creation for TWAs

Dp-Goog:
The earlier changes worked to create the high priority notification channel and delete the old notification channel if that existed. More investigation revealed that it was flaky, and the root cause is that the super class for AndroidX always posted to the default channel, even though it seemed like the code in NotificationUtils was actually creating the correct channel.

This PR fixes that by overriding `onNotifyNotificationWithChannel()` by intercepting the notification, and manually setting a a high priority channel if the correct metadata exists. Verified manually using the test TWA that it works, with additional logging.


Dp-Goog:
@PEConn or @gdziadosz-google or @dtonderski-google , can anyone review this PR and help land this? This should be the last blocking thing needed for TWAs to support high priority notifications, thanks!

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -78,4 +85,38 @@ public Bundle onExtraCommand(
...
+            NotificationManager notificationManager =
+                    (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                NotificationChannel channel = notificationManager.getNotificationChannel(channelId);
+                if (channel != null && channel.getImportance() == NotificationManager.IMPORTANCE_NONE) {
+                    return false;
+                }
+
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
```
gdziadosz-google: If I'm checking correctly, `Build.VERSION_CODES.O` is API 26 and `Build.VERSION_CODES.N` is API 24. In that case the inner version check is redundant.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -78,4 +85,38 @@ public Bundle onExtraCommand(
...
     }
+
+    @Override
+    @RequiresPermission(android.Manifest.permission.POST_NOTIFICATIONS)
+    public boolean onNotifyNotificationWithChannel(
+            @NonNull String platformTag,
+            int platformId,
+            @NonNull Notification notification,
+            @NonNull String channelName) {
+        if (NotificationUtils.shouldUseHighPriorityNotifications(this)) {
```
PEConn: nit: Turn this into a guard clause:

```
if (!shouldUseHighPriorityNotification) {
    return super.onNotifyNotificationWithChannel();
}

// the rest
```

PEConn (approved):
LGTM % nit

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
Dp-Goog: Aaah yes, fixed.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
Dp-Goog: Done!

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -78,4 +85,35 @@ public Bundle onExtraCommand(
     public void registerExtraCommandHandler(ExtraCommandHandler handler) {
         mExtraCommandHandlers.add(handler);
     }
+
+    @Override
+    @RequiresPermission(android.Manifest.permission.POST_NOTIFICATIONS)
+    public boolean onNotifyNotificationWithChannel(
```
gdziadosz-google, line 92: Looking at the implementation in the TrustedWebActivityService class, looks like it was posting the notification on the default channel because it's using its own `channelNameToId` method. That method is also used in `onAreNotificationsEnabled` - should that one be also overridden?

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -78,4 +85,35 @@ public Bundle onExtraCommand(
...
+
+        NotificationUtils.createNotificationChannel(this, channelName);
+        String channelId = NotificationUtils.channelNameToId(this, channelName);
+
+        NotificationManager notificationManager =
+                (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            NotificationChannel channel = notificationManager.getNotificationChannel(channelId);
+            if (channel != null && channel.getImportance() == NotificationManager.IMPORTANCE_NONE) {
```
gdziadosz-google: Super class also checks at the beginning of this method if notifications are enabled. Here it's not checked. Maybe you should call here `NotificationUtils.areNotificationsEnabled` instead of checking only the channel's importance level?

Dp-Goog: Done

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
Dp-Goog: Fwiw, I was able to get the notifications running even without these changes applied (shared video on DM), but still made these changes because it seems more robust that way.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
gdziadosz-google: Condition from this if statement (`channel != null && channel.getImportance() == NotificationManager.IMPORTANCE_NONE`) is also checked in `NotificationUtils.areNotificationsEnabled`, so calling this method can replace this `if` (that's why I suggested to call it instead of checking importance level here).

Right now we have 2 if statements where the first one is redundant. So replace:
```
if (channel != null && channel.getImportance() == NotificationManager.IMPORTANCE_NONE)
```
with
```
if(!NotificationUtils.areNotificationsEnabled(this, channelName))
```

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
```diff
@@ -93,27 +95,46 @@ public boolean onNotifyNotificationWithChannel(
             int platformId,
             @NonNull Notification notification,
             @NonNull String channelName) {
+        Log.i(TAG, "onNotifyNotificationWithChannel called. tag=" + platformTag + ", id=" + platformId + ", channelName=" + channelName);
```
gdziadosz-google: Let's remove all those log statements (I assume they were added temporarily for debugging purposes).

gdziadosz-google (changes requested):

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
Dp-Goog: Aaaaah, fixed them.

androidbrowserhelper/src/main/java/com/google/androidbrowserhelper/trusted/DelegationService.java
Dp-Goog: Aah yes, done.

gdziadosz-google (approved):

PR #618, opened 2026-08-24
Remove @VisibleForTesting annotations in notifications code

Dp-Goog:
The default visibility in package-private, so tests should have access to these functions anyway.


Dp-Goog:
@gdziadosz-google or @gstepniewski-google PTAL, thanks!

gdziadosz-google (approved):

PR #621, opened 2026-08-27
Ensure notification popups do not lag behind notification sounds, mimicing "appear as popup" on the settings

Dp-Goog:
Most notifications have the group alert behavior set to `GROUP_ALERT_SUMMARY`, which hides the popups behind the sound of the notification coming in the shade to show a single notification popup, giving the "feeling" that the notification is lagging.

Fix that by ensuring that `GROUP_ALERT_ALL` for TWA notifications so that popups appear instantly, bringing behavior closer to other desktop platforms.


Dp-Goog:
@PEConn or @gdziadosz-google or @dtonderski-google can you please help review this? Thanks!

gstepniewski-google (approved):

